PR #1, opened 2015-10-02
Report physical blocks used

andrewgaul:
goofys does not support sparse files so always report the object size
for files.


andrewgaul:
Do not merge until upstream merges jacobsa/fuse#12.

andrewgaul:
Not needed as fuse library now forces the physical size to be the same as the logical size.

PR #13, opened 2015-10-15
Simplify usage for credentials

andrewgaul:
Give the user the quick and dirty example of a simple access key and
credential and point them at the AWS CLI for more complicated uses.


kahing:
merged via 4d255ca96532a3b0017ad00bf947561db7f2c809

PR #43, opened 2015-11-18
test the fuse layer

kahing:
fuse-test.sh is mostly from s3fs, with some minor modifications
to make it run in this environment and some tests disabled

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,12 @@
 language: go
-sudo: false
+sudo: required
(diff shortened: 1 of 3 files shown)


andrewgaul:
Add the following to `travis.yml`:

```
sudo: required
dist: trusty
```

See s3fs-fuse/s3fs-fuse@94e3dbb2dc3e6de253d2313a4d0848e61fd46803 for an example.

Issue #84, opened 2016-06-08
Enable use of InstanceProfile on EC2 instances


robparrott:
On EC2 instances with an InstanceProfiles the binary should use the provided IAM temp user capability provided by AWS when an explicit profile is not set.

See http://docs.aws.amazon.com/sdk-for-go/api/aws/Config.html#Credentials-field For details.

robparrott:
Expect a PR shortly for this.

PR #86, opened 2016-06-08
restoring default credentials object

jfwarner:
Fix for #84

Restoring default credentials object to "a chain of credential providers to search for credentials in environment variables, shared credential file, and EC2 Instance Roles" 

http://docs.aws.amazon.com/sdk-for-go/api/aws/Config.html#Credentials-field

Previously, the "Credentials" field was left nil if there was no "--profile" argument. Now the Credentials field will be left at the default.

### DIFF
--- a/main.go
+++ b/main.go
@@ -73,17 +73,16 @@ func mount(
 	mountPoint string,
 	flags *FlagStorage) (mfs *fuse.MountedFileSystem, err error) {
 
-	var creds *credentials.Credentials
-	if len(flags.Profile) > 0 {
-		creds = credentials.NewSharedCredentials(os.Getenv("HOME")+"/.aws/credentials", flags.Profile)
-	}
-
 	awsConfig := &aws.Config{
 		Region:      &flags.Region,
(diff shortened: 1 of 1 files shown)


kahing:
I was always under the assumption that the default value for a struct pointer is `nil`, is that not the case?

jfwarner:
I am a Go novice, so I don't quite understand everything that is going on yet. But this documentation leads me to believe otherwise:
http://docs.aws.amazon.com/sdk-for-go/api/aws/Config.html#Credentials-field

I also tested this change, and it behaves exactly the way the above documentation describes.

jfwarner:
Perhaps it is the case that a nil value induces the desired behavior in the SDK, as in:
if( field is nil) then { use default chain };

In either case, it appears to work.

kahing:
I found this difficult to believe because if --profile is not used, `awsConfig.Credentials` is nil before and after the patch.

I will accept this anyway because it makes the config handling more consistent with other overrides.

PR #91, opened 2016-06-25
Add a --credentials option and respect AWS environment variables

blampe:
This resolves an edge case where we were unable to specify a custom credentials path with `AWS_SHARED_CREDENTIALS_FILE` while also specifying `--profile`.

Before, on a system where `$HOME/.aws/credentials` doesn't exist but `~/.aws/custom-creds` is available:
```
$ AWS_SHARED_CREDENTIALS_FILE=~/.aws/custom-creds goofys -f --profile foo bucket ~/s3
2016/06/24 22:19:38.749413 main.ERROR Unable to access 'bucket': invalid argument
2016/06/24 22:19:38.749664 main.FATAL Mounting file system: Mount: initialization failed
```

Now:
```
$ AWS_SHARED_CREDENTIALS_FILE=~/.aws/custom-creds goofys -f --profile foo bucket ~/s3
2016/06/24 22:18:36.171359 main.INFO File system has been successfully mounted.
```

As a side-effect this is more portable to Windows, too, I guess. See https://github.com/aws/aws-sdk-go/blob/master/aws/credentials/shared_credentials_provider.go#L27 for more detail.

Also adds a `--credentials` option to specify a custom path and extends `.gitignore` to be a little more vim-friendly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,33 @@
+# Editor files
 *~
+*.swp
+
+# Binaries
 goofys
-goofys.test
 xout
 s3proxy.jar
+
+# Compiled Object files, Static and Dynamic libs (Shared Objects)
+*.o
+*.a
+*.so
+
+# Folders
+_obj
+_test
+
+# Architecture specific extensions/prefixes
+*.[568vq]
+[568vq].out
+
+*.cgo1.go
+*.cgo2.c
+_cgo_defun.c
+_cgo_gotypes.go
+_cgo_export.*
+
+_testmain.go
+
+*.exe
+*.test
+*.prof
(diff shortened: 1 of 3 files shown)


kahing:
Instead of all that changes, what if we just do this:

```golang
diff --git a/main.go b/main.go
index 9ef41ec..fef9122 100644
--- a/main.go
+++ b/main.go
@@ -74,13 +74,13 @@ func mount(
 
        if len(flags.Profile) > 0 {
-               awsConfig.Credentials = credentials.NewSharedCredentials(os.Getenv("HOME")+"/.aws/credentials", flags.Profile)
+               awsConfig.Credentials = credentials.NewSharedCredentials("", flags.Profile)
        }
 
        if len(flags.Endpoint) > 0 {
```

PR #92, opened 2016-06-25
Also expire inode cache on unlink

blampe:
This should resolve an issue where files couldn't be moved/copied from within the GUI in OS X. Similar to #88.

`mv` and `cp` worked fine for me but when I tried to do similar operations from the Finder I consistently got errors about the file already existing ("The operation can’t be completed because an item with the name “some-file copy” already exists").

I suspected this was due to a caching problem and to confirm I set TTLs to 0 and was able to successfully move the files around.

I looked into the fuse logs and (as far as I can tell) in order to duplicate a file the Finder will first create a 0-byte placeholder at the path it intends to write to, unlink it, and then actually write contents to that path. We weren't clearing the inode cache on the unlink step, so this caused the Finder to think it was overwriting something.

### DIFF
--- a/internal/goofys.go
+++ b/internal/goofys.go
@@ -932,6 +932,12 @@ func (fs *Goofys) Unlink(
 	fs.mu.Unlock()
 
 	err = parent.Unlink(fs, op.Name)
+	if err == nil {
+		fs.mu.Lock()
+		fullName := parent.getChildName(op.Name)
+		delete(fs.inodesCache, fullName)
+		fs.mu.Unlock()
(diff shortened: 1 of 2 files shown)


kahing:
I merged the first commit but not the run-tests changes. stdbuf allows s3proxy's debug output to be in sync with goofys' and is sometimes helpful when there's a bug. Maybe you can install stdbuf for OS X? https://apple.stackexchange.com/questions/193141/to-use-stdbuf-from-homebrews-coreutils

PR #106, opened 2016-09-09
Feature/sse s3

robparrott:
Adds a new `--use-sse` option that makes all uploads set with AES-256 server encryption.

Implements issue #106

### DIFF
--- a/internal/goofys.go
+++ b/internal/goofys.go
@@ -462,6 +462,7 @@ func (fs *Goofys) copyObjectMultipart(size int64, from string, to string, mpuId 
 			Key:          fs.key(to),
 			StorageClass: &fs.flags.StorageClass,
 			ContentType:  fs.getMimeType(to),
+			ServerSideEncryption: &fs.flags.SSEType,   
 		}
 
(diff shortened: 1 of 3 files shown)


internal/goofys.go
```diff
@@ -462,6 +462,11 @@ func (fs *Goofys) copyObjectMultipart(size int64, from string, to string, mpuId
 			Key:          fs.key(to),
 			StorageClass: &fs.flags.StorageClass,
 			ContentType:  fs.getMimeType(to),
+			ServerSideEncryption:  &fs.flags.SSEType,
+		}
+
+		if  fs.flags.UseSSE  {
```
kahing: Not at a computer now but this seems redundant?

kahing:
Could you squash this?

internal/goofys.go
robparrott: Yup. My mistake. Good eye!

robparrott:
Changed.

kahing:
55c235cf0dec33bd5c0542e5c0ed4fe2ced7628f

Issue #102, opened 2016-09-02
Allow for use of SSE-KMS encryption with a specified KMS Key ARN


robparrott:
We'd like to be able to mount a goofy volume with parameters that specify that all writes are encrypted using a specified KMS key (provided by ARN).

It also needs to be able to read any object from S3 that are SSE-KMS encrypted with the specified KMS key.

robparrott:
This feature would be a follow up to #105 since it involves setting the appopriate header to "aws:kms" and optionally setting a KMS key ARN id.

robparrott:
PR Submitted: https://github.com/kahing/goofys/pull/107

kahing:
55c235cf0dec33bd5c0542e5c0ed4fe2ced7628f

robparrott:
Thanks for merging this so fast!

Any chance of cutting a new release with this set of new features and fixes?

kahing:
only if you guys tell me what you are up to with goofys ;-)

robparrott:
Sure,

We're using it to provide volume mounts to docker containers so that import/export and diagnostics capture will use a centralized object store, and not a local or shared filesystem. In the future we plan to add features to the product to write directly to S3 for some of this, but for now this provides a technology bridge.

PR #107, opened 2016-09-09
Feature/sse kms

robparrott:
For Issue https://github.com/kahing/goofys/issues/102

This builds on the PR for SSE-S3.

Adds two more options: `--use-kms` and `--kms-key-id` with expanded options for server side encryption of objects in S3. 

* Requires that `--use-sse` is specified. 
* When only  the `--use-kms` is given, then uses the KMS server-side encryption mode (SSE-KMS) but with the default, customer master key (CMK).
* When `--kms-key-id` is specified in addition to `--use-kms` it use use that specified KMS instead of the master key. Requires that the user that is connected to S3 via goofys has permission to access and use that KMS key.

All three cases tested in  basic fashion with small and large file sizes.

### DIFF
--- a/internal/flags.go
+++ b/internal/flags.go
@@ -262,7 +262,8 @@ func PopulateFlags(c *cli.Context) (flags *FlagStorage) {
 		UsePathRequest: c.Bool("use-path-request"),
 		Profile:        c.String("profile"),
 		UseContentType: c.Bool("use-content-type"),
-		UseSSE:         c.Bool("use-sse"),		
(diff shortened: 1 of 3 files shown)


kahing:
I squashed this and #106 and massaged it a little. Could you take a look and let me know what you think?

robparrott:
This makes sense given that the --sse-kms string can be left as empty.

kahing:
merged via 55c235cf0dec33bd5c0542e5c0ed4fe2ced7628f

PR #121, opened 2016-10-13
Adds a new --acl option that allows to set the canned ACL to be applied to the object

monken:
Default is off. Allowed values are private | public-read |
public-read-write | authenticated-read | aws-exec-read | bucket-owner-read
| bucket-owner-full-control.

### DIFF
--- a/internal/flags.go
+++ b/internal/flags.go
@@ -154,6 +154,13 @@ func NewApp() (app *cli.App) {
 				Value: "",
 			},
 
+			/// http://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview.html#canned-acl
+			cli.StringFlag{
+				Name:  "acl",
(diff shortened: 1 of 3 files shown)


monken:
@kahing I just rebased it off of master so it should merge cleanly.

kahing:
should this be --canned-acl or --acl-canned instead of just --acl? also, what do you think about prefixing aws specific options with --aws-blah? There are many S3 implementations that don't support all these aws features and grouping them may make --help easier to read (or at least group them together in a different section)

Also, could you change the possible values display to follow the convention in the other flags? (`Possible values: a, b, c` instead of `allowed values: a | b | c`)

monken:
I agree that the `aws` prefix could make it cleaner. However, the `sse` and `sse-kms` flags are not supported by services other than S3 either. I tried to be consistent with the implementation of that feature. 

It's your call, I'm happy to change it :)

monken:
> Also, could you change the possible values display to follow the convention in the other flags? (Possible values: a, b, c instead of allowed values: a | b | c)

Done

kahing:
merged as ac4f3b6769dc31f4a319e9a0ee6fbf8e2e947717 . amended help text

Issue #122, opened 2016-10-14
Unable to create directories if bucket enforces server side encryption


dynamike:
If a bucket enforces that all PUTs have server side encryption enabled via a bucket policy then I'm unable to create directories. 

```
./goofys -o allow_other --uid <uid> --gid <uid> --sse -o allow_other <bucket-name> /test-mount
root@instance:/test-mount# touch test
root@instance:/test-mount# rm test
root@instance:/test-mount# mkdir dir
mkdir: cannot create directory ‘dir’: Permission denied
```
Now if I remove the bucket policy that is enforcing server side encryption it works
```
root@instance:/test-mount# mkdir ho
```

kahing:
fixed by #123

PR #123, opened 2016-10-14
Use SSE(if enabled) when calling MkDir

dynamike:
This fixes the issues I saw in #122


kahing:
great thanks! brownie points for whoever refactors this so we don't add this everywhere we do PUT (relevant for #121 as well).

PR #134, opened 2016-11-29
Add mnt-mode option to support chroot usage

sspans:
This makes mnt mode a seperate option which is useful
to allow sftp/chroot to a goofys mount. By default a 'writeable'
goofys mount and 'ChrootDirectory %h' will cause sshd/sftpd to fail
with the following message:
'fatal: bad ownership or modes for chroot directory'

Running with --mnt-mode=755 makes chrooted sftp possible.

For more details see:
http://serverfault.com/questions/656753/configuring-chroot-for-sftp-users
https://github.com/s3fs-fuse/s3fs-fuse/pull/110

### DIFF
--- a/internal/flags.go
+++ b/internal/flags.go
@@ -92,6 +92,12 @@ func NewApp() (app *cli.App) {
 			},
 
 			cli.IntFlag{
+				Name:  "mnt-mode",
+				Value: 0755,
+				Usage: "Permission bits for the mount point. (default: 0755)",
+			},
+			
+			cli.IntFlag{
 				Name:  "dir-mode",
 				Value: 0755,
 				Usage: "Permission bits for directories. (default: 0755)",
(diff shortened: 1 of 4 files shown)


kahing:
should goofys just use the mountpoint directory permission before the mount as the permission instead of using another option?

sorry I haven't been able to look at this earlier. Note that I will be traveling for the next two weeks and responses maybe sparse.

sspans:
The idea is to have different permissions for the mountpoint versus the directories below that.
Making it configurable seemed like the cleanest solution.

kahing:
Wouldn't `--mnt-mode=755` just be the same as not using that option as that's the same as the default `dir-mode`?

I really don't like to have too many configuration options, could you try making it use the mountpoint's original attributes for the root mode, uid, and gid? That way it's easier to satisfy

> At session startup sshd(8) checks that all components of the pathname are root-owned directories which are not writable by any other user or group

without making everything root as well

PR #148, opened 2017-01-29
Add Homebrew instructions

andrewgaul:
Also break out installation and usage.


kahing:
I don't know anything about homebrew, but is it possible to make goofys depend on fuse so the user doesn't need two commands?

andrewgaul:
Unfortunately Homebrew hosts FUSE formulae in a separate repository from others.

kahing:
homebrew PR here: https://github.com/Homebrew/homebrew-fuse/pull/64

PR #151, opened 2017-02-16
UPDATE README.md filey system TO file system

sungjunyoung:
I do not know if it's a typo or an intent, but I think it's a good idea to write 'file system' and send a pull request.

Thank you for making good open source.


sungjunyoung:
NOOP mistake. I have not read README.md properly.... sorry

Issue #161, opened 2017-03-07
SIGSEGV when running parallel du on large goofys directory


andrewgaul:
```
$ go run du/du.go /path/to/linux-4.10/ > /dev/null
du: readdirent: software caused connection abort
du: readdirent: software caused connection abort
du: readdirent: software caused connection abort
```

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x7b4892]

goroutine 1926 [running]:
github.com/kahing/goofys/internal.(*DirHandle).ReadDir(0xc421762380, 0xc4200dedc0, 0x2, 0xc42030a6a0, 0x2, 0x2)
        /home/gaul/work/go-workspace/src/github.com/kahing/goofys/internal/handles.go:1223 +0x912
github.com/kahing/goofys/internal.(*Goofys).ReadDir(0xc4200dedc0, 0x7f396654b090, 0xc42163d860, 0xc4217348c0, 0x0, 0x0)
        /home/gaul/work/go-workspace/src/github.com/kahing/goofys/internal/goofys.go:904 +0x1ec
github.com/jacobsa/fuse/fuseutil.(*fileSystemServer).handleOp(0xc42017a200, 0xc420122240, 0x7f396654b090, 0xc42163d860, 0x7f2820, 0xc4217348c0)
        /home/gaul/work/go-workspace/src/github.com/jacobsa/fuse/fuseutil/file_system.go:164 +0xa23
created by github.com/jacobsa/fuse/fuseutil.(*fileSystemServer).ServeOps
        /home/gaul/work/go-workspace/src/github.com/jacobsa/fuse/fuseutil/file_system.go:108 +0x12a
```

andrewgaul:
Extra logging reveals an object without a Size:

```
{
  ETag: "\"0700e5a8ffa2b7a6722f24fb16569a05\"",
  Key: "linux-4.10/arch/tile/kernel/sysfs.c",
  LastModified: 2017-03-07 10:40:38 +0000 UTC
}
```

kahing:
does this object always not have a size?

andrewgaul:
Different errors on different runs.  Now I encountered:

```
panic: runtime error: slice bounds out of range

goroutine 64772 [running]:
[...]

Issue #157, opened 2017-02-26
need to update aws-sdk-go for golang1.8


kahing:
https://aws.amazon.com/blogs/developer/aws-sdk-for-go-update-needed-for-go-1-8/

kahing:
fixed by 044cae614c21fef469de173797c26ca603b22292

PR #172, opened 2017-04-13
add nil check

arc279:


andrewgaul:
@arc279 Can you share your motivations for this PR?  I encountered similar symptoms in #161 but upgrading the SDK in #157 might have addressed them.

kahing:
I suspect this PR is marking the problem and the real issue is https://github.com/aws/aws-sdk-go/issues/1205 . Marking this as dup of #161

PR #190, opened 2017-06-06
add --no-dir-compat option to minimize LIST requests

jtwang83:
add --no-dir-compat option like s3fs's notsup_compat_dir to minimize LIST requests at the expense of compatibility

In my deployment, I have noticed an enormous number of expensive Tier1 requests. From the goofys debug logs, I traced this to redundant checks for directories not created by goofys in LookUpInodeMaybeDir. By short circuiting these checks, we can eliminate a LIST request when opening a new file.

Note this is my first time working with go, however the change is very minor and passes the tests with the cross product of options --cheap and --no-dir-compat.

### DIFF
--- a/internal/flags.go
+++ b/internal/flags.go
@@ -184,6 +184,11 @@ func NewApp() (app *cli.App) {
 				Usage: "Reduce S3 operation costs at the expense of some performance (default: off)",
 			},
 
+			cli.BoolFlag{
+				Name:  "no-dir-compat",
+				Usage: "Assume all directory objects are \"dir/\" (default: off)",
+			},
+
 			cli.DurationFlag{
 				Name:  "stat-cache-ttl",
 				Value: time.Minute,
@@ -229,7 +234,7 @@ func NewApp() (app *cli.App) {
 		flagCategories[f] = "aws"
 	}
 
-	for _, f := range []string{"cheap", "stat-cache-ttl", "type-cache-ttl"} {
+	for _, f := range []string{"cheap", "no-dir-compat", "stat-cache-ttl", "type-cache-ttl"} {
 		flagCategories[f] = "tuning"
 	}
 
@@ -264,6 +269,7 @@ type FlagStorage struct {
 
 	// Tuning
 	Cheap        bool
+	NoDirCompat	 bool
 	StatCacheTTL time.Duration
 	TypeCacheTTL time.Duration
 
@@ -309,6 +315,7 @@ func PopulateFlags(c *cli.Context) (flags *FlagStorage) {
 
 		// Tuning,
 		Cheap:        c.Bool("cheap"),
+		NoDirCompat:  c.Bool("no-dir-compat"),
 		StatCacheTTL: c.Duration("stat-cache-ttl"),
 		TypeCacheTTL: c.Duration("type-cache-ttl"),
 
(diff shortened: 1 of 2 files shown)


kahing:
if you short circuit LIST for directories without the directory blob, wouldn't you just ignore those directories? Under what circumstance would this be a good thing?

seems like the right thing to do is fix those directories to have the right blobs

jtwang83:
The idea is that if you know you only ever wrote to the bucket with goofys (which creates the blobs), it is safe to short circuit the LIST.

If you tried to fix the directories to have the right blobs upon access, you would still have to do the LIST to determine if they are there. If you have directories that don't have the blobs and you want to minimize your LIST requests, you'd need to run a separate program to scan through the bucket to create the blobs, then commit to only creating dirs using goofys in the future.

kahing:
right, I see what you mean. what if we store the number of directory blobs and number of times we have to list key/, and use some heuristics to skip listing after some point? That seems to be the better thing to do and we can add --always-list when/if people complain.

jtwang83:
I think that's OK, but I don't think we need that level of complexity. If a user knows that they aren't writing to the bucket with other tools, then they should use this option; otherwise they shouldn't.

To put it another way, why not apply the same heuristic to the --cheap option?

kahing:
if you can suggest better heuristic for --cheap so that goofys can determine automatically what to do, I am all ears.

Sorry for being a pain, it's my duty to push back extra options if the software can figure out what The Right Thing to do is.

jtwang83:
I think a heuristic for --cheap makes more sense than a heuristic for --no-dir-compat. With --cheap, if goofys changes behavior based on the heuristic without a change in the invocation options, the impact is in performance but the ultimate functionality is the same.

A heuristic for --no-dir-compat would be very confusing because a directory without the blob would be visible sometimes but not at other times. Imagine a user mounting a bucket with a long running process. Everything works great. Then the user puts a file in a directory with a tool that doesn't create the blob, and goofys will never see it because the heuristic has been satisfied.

kahing:
I am still a little confused about why this is necessary. If the directory blobs are there, and you are using --cheap, then goofys only does LIST if you access a path that doesn't exist. 

Ah I see, you are creating new files and hence there are lots of lookups for paths that don't exist. In that case, there will be one LIST for each new file. Depending on how large the file is, this could be 50% of the operation cost (one LIST and one PUT for small files).

jtwang83:
Correct, the issue is with small files.

Somewhat unrelated, it would be useful to scale partSize() with the available memory on the box. Large files take a large number of PUT requests as the first 5G are uploaded 5MB at a time.

internal/flags.go
```diff
@@ -184,6 +184,11 @@ func NewApp() (app *cli.App) {
 				Usage: "Reduce S3 operation costs at the expense of some performance (default: off)",
 			},
 
+			cli.BoolFlag{
+				Name:  "no-dir-compat",
```
kahing: gcsfuse uses --implicit-dirs for the inverse behavior and that's the closest I could find, maybe call this --explicit-dirs or --no-implicit-dirs for consistency?

internal/flags.go
```diff
@@ -184,6 +184,11 @@ func NewApp() (app *cli.App) {
...
+				Usage: "Assume all directory objects are \"dir/\" (default: off)",
```
kahing: I think this reads better:

Assume all directory objects ("dir/") exist (default: off)

kahing:
thanks for the contribution @jtwang83 ! I amended and pushed the commit with the changes I listed out above

jtwang83:
Thanks! I got swamped so I abandoned this, but thanks for picking up the slack.

PR #191, opened 2017-06-06
Lowercase github.com/Sirupsen/logrus

robmadole:
From https://github.com/sirupsen/logrus#logrus-

> Seeing weird case-sensitive problems? Unfortunately, the author failed to realize the consequences of renaming to lower-case. Due to the Go package environment, this caused issues. Regretfully, there's no turning back now. Everything using logrus will need to use the lower-case: github.com/sirupsen/logrus. Any package that isn't, should be changed.

### DIFF
--- a/internal/buffer_pool_test.go
+++ b/internal/buffer_pool_test.go
@@ -20,7 +20,7 @@ import (
 	"sync"
 	"time"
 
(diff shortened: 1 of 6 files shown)


kahing:
thanks for the contribution!

PR #198, opened 2017-07-06
update homebrew instruction

khoan:
homebrew/fuse has been merged into homebrew core: https://github.com/Homebrew/homebrew-core/pull/11283


kahing:
thanks @khoan ! I don't use OS X so contributions like this is very valuable.

PR #213, opened 2017-08-02
URL encode ListObject responses

gaul:
This allows file names with ASCII values 1-10.


andrewgaul:
S3Proxy seems to have a bug when using `encoding-type`.

andrewgaul:
Must wait for the next S3Proxy version before merging this.

gaul:
This pull request depends on #295.

Issue #216, opened 2017-08-22
panic: double insert of ... when moving one file to another.


jonykalavera:
this happens when executing a mv from one file to another. ej `mv new.jpg new`
```
2017/08/22 18:11:25.770046 s3.INFO Switching from region 'us-east-1' to 'us-west-2'
2017/08/22 18:11:26.184980 main.INFO File system has been successfully mounted.
panic: double insert of filer_public/27/40/27400712-4eab-40ee-84dc-9869b98919fe/new

goroutine 53648 [running]:
github.com/kahing/goofys/internal.(*Inode).insertChildUnlocked(0xc420d15dc0, 0xc422003a40)
	/home/khc/Code/go/src/github.com/kahing/goofys/internal/handles.go:267 +0x626
github.com/kahing/goofys/internal.(*Goofys).Rename(0xc420099320, 0x7f06016fccb8, 0xc42084a2d0, 0xc42084a2a0, 0x0, 0x0)
	/home/khc/Code/go/src/github.com/kahing/goofys/internal/goofys.go:1444 +0x38b
github.com/jacobsa/fuse/fuseutil.(*fileSystemServer).handleOp(0xc420224260, 0xc420530180, 0x7f06016fccb8, 0xc42084a2d0, 0x8c3c00, 0xc42084a2a0)
	/home/khc/Code/go/src/github.com/jacobsa/fuse/fuseutil/file_system.go:156 +0x226
created by github.com/jacobsa/fuse/fuseutil.(*fileSystemServer).ServeOps
	/home/khc/Code/go/src/github.com/jacobsa/fuse/fuseutil/file_system.go:112 +0x12a
```

kahing:
what version is this? I cannot reproduce this

kahing:
please reopen if you can reproduce this with the last release or master

inevity:
I also met this issue .When update to master, the panic disappear.
But when repeat mv a b ,second,b have existing, will panic on the third mv.
please ref: https://github.com/kahing/goofys/pull/225/commits/e8ae74ca5e3a9283d322f0b441aa7bc54bd1b16d.

PR #225, opened 2017-09-19
removechild fix when rename existing file repeately

inevity:
The issue will lead to removeChildUnlocked panic.

It relate the issue https://github.com/kahing/goofys/issues/216.


kahing:
could you write a test for this also? That way this won't accidentally happen again

kahing:
thanks for finding this bug, I don't think your fix is correct but I've just checked in a proper fix.

inevity:
In essense, the issue is relate with relationship between inode and dentry(in goofys,call child).  one inode may have multiply dentries ref . such as rename ,we remove child ,we should not remove all dentry relate with the inode .Instead, we should remove the dentry relate the name rename op supplied to.

PR #227, opened 2017-09-23
readirop should check whether reach end first before readdir from cache

inevity:
When first readdir offset from s3 backend, then second readdir offset from userspace will lead goofys readdir from cache.
if the first readdir have read all direntries,then second will go through cache. In some case, will lead to two same entry return to user.


inevity:
For commit 4284dc4, when dir fd open in a shell,rm -rf this dir in another shell.In fuse ,only when fd close/release, kernel will send forgetinode to goofys. but with normal rm , ,kernel will send forgetinode following the unlink/rm immediately
So if we not invalidate the goofys inode ,the next lookup from (kernel invadation or  stat,open,etc) will found inode in goofys. but this is not expected.

The test case will be appended .

kahing:
have you been able to add a test case?

inevity:
Add test_rmdir_withfdopen to fuse-test.sh;goofystest
    TestRmdirWithDiropen;standalone test script for rmwithopen
    note:
    standalone shell test  scripts is located in the test dir,it's same
    as the test_rmdir_withfdopen in the fuse-test.sh. The standalone
    script and fuse-test testcase use a binary program to open dir and
    sleep for a long time .
    
    Gooystest TestRmdirWithDiropen use the go exec to test.
    
    At present, only the standalone test  scripts can trigger this bug.
    the fuse_test and goofystest testcase cannot trigger which i have
    not found why.


When use the standalone shell test scripts,
patch before:
sh /root/deploy/test/lsun/testrmwithopen.sh
/root/deploy/test/lsunconsistency: no process found
ls /mnt/s3
total 8.0K
drwxr-xr-x. 2 root root 4.0K 11月  1 16:14 .
drwxr-xr-x. 9 root root   99 9月   8 12:09 ..
drwxr-xr-x. 2 root root 4.0K 11月  1 16:14 test111
ls /mnt/s3/test111
test1 //note: two same dentry
test1
ls -lah /mnt/s3/test111
total 16K
drwxr-xr-x. 2 root root 4.0K 11月  1 16:14 .
drwxr-xr-x. 2 root root 4.0K 11月  1 16:14 ..
drwxr-xr-x. 2 root root 4.0K 11月  1 16:14 test
drwxr-xr-x. 2 root root 4.0K 11月  1 16:14 test1
should only test1 dir


after patch:
/root/deploy/test/lsunconsistency: no process found
ls /mnt/s3
total 8.0K
drwxr-xr-x.  2 root root 4.0K 11月  1 17:00 .
drwxr-xr-x. 10 root root  112 11月  1 16:32 ..
drwxr-xr-x.  2 root root 4.0K 11月  1 17:00 test111
ls /mnt/s3/test111
test1
ls -lah /mnt/s3/test111
total 12K
drwxr-xr-x. 2 root root 4.0K 11月  1 17:00 .
drwxr-xr-x. 2 root root 4.0K 11月  1 17:00 ..
drwxr-xr-x. 2 root root 4.0K 11月  1 17:00 test1
should only test1 dir

test/fuse-test.sh
```diff
@@ -453,6 +453,40 @@ function test_write_after_seek_ahead {
...
+    
+    mkdir -p  $TEST_BUCKET_MOUNT_POINT_1/$TEST_DIR/test
+    mkdir -p  $TEST_BUCKET_MOUNT_POINT_1/$TEST_DIR/test1
+    echo "ls mountpoint"
+    lsout0=`ls -lah $TEST_BUCKET_MOUNT_POINT_1/`
+    echo $lsout0
+
+
+
+   nohup /root/go/src/github.com/kahing/goofys/test/rmdirwihtopen/lsunconsistency $TEST_BUCKET_MOUNT_POINT_1/$TEST_DIR/test a > /dev/null 2>&1 &
```
kahing: what is this lsunconsistency?

internal/goofys_test.go
```diff
@@ -1956,3 +1956,73 @@ func (s *GoofysTest) TestRead403(t *C) {
 	_, err = fh.ReadFile(0, buf)
 	t.Assert(err, Equals, syscall.EACCES)
 }
+
+func (s *GoofysTest) TestRmdirWithDiropen(t *C) {
```
kahing: this test passes with unmodified goofys

test/fuse-test.sh
kahing: ah found that you committed a binary. Please replace that with the source (maybe C) file instead. I am not going to run a random binary to verify the problem

internal/goofys_test.go
kahing: I see, that's because in goofys_test the tests are by default with with 0 type cache ttl and stat cache ttl. Once I enable those I can make the test fail

kahing:
I checked in a fix with your test, slightly modified. Thanks for the PR! I don't understand 
ff2421c though, is that a different issue and could you submit a test that shows the problem?

inevity:
Have checked the pathc

backend | standshelltest | goofys_test|
------------ | ------------- | ------------|
s3proxy| passed|passed|
minio  |NO TO TEST |failed|
s3like|passed|passed|

Need to dig into the failed.
As for the ff2421c ,maybe should fake a testcase?!

inevity:
Minio now does not support empty dir such as dir/,even return 200 OK.

inevity:
The behavior of minio and aws s3  treat dir/ is not the object in  durable storagebackend .The dir object is only a concept parsed from the object name which have slash such as dir/dir1/file1.

In Minio, when put dir/, minio server will mkdir dir,return ok.
When list dir/, minio server will readdir this empty dir, return ok,but have nothing to return. 
IN goofys func isEmptyDir , len(resp.Contents)  should equals 0.

This behavior is different with s3proxy which will len(resp.Contents) = 1 ,including the dir/ as a key when list this empty dir.

Anothe found:In minio,for nonexist dir/, list dir/ will return errr,(nu such file or diretory,here is dir/)
                                                               Head dir/, same errror.
                       In minio,if we delete file or dir, it will recursilly delete empty parent dir. .For example,
                       a of dir  only has one file b. Since we delete a/b, if a is empty, it will also delete a.

If we align the rule that dir/ is not the object in  durable storagebackend,we should insist that the result of list dir should not have the dirname as the key in the Contents' section.  At least, chang the design of minio is difficult now,
so we should treat minio diffrently as s3proxy, or we should treat them unitedly(differently?).



What is your idea? 
@kahing @gaul

kahing:
minio is not fully supported because last I checked, it does not support objects ending with a / (slash). If you could rebase this PR and add a test case for https://github.com/kahing/goofys/commit/ff2421c8f1bbd56b7b3291d25f8137dbd3881e87 I will take a look again

PR #243, opened 2017-10-19
fix readahead panic when s3 return 'permission denied'

inevity:
If getobject req return err such as permission denied, the readaheader buffer Buffer struct member reader will be nil, member err = 13(permission denied).but member buf(type Mbuf) is not nil.

So when do Buffer.Read,first readfrom buf,then read from reader,which lead to panic.because reader is nil.

### DIFF
--- a/internal/buffer_pool.go
+++ b/internal/buffer_pool.go
@@ -377,7 +377,9 @@ func (b *Buffer) Read(p []byte) (n int, err error) {
 		b.cond.Wait()
 	}
 
-	if b.buf != nil {
+	if b.err != nil && b.err != io.EOF {
+		err = b.err
+	} else if b.buf != nil {
 		bufferLog.Debugf("reading %v from buffer", len(p))
 
 		n, err = b.buf.Read(p)
@@ -389,8 +391,6 @@ func (b *Buffer) Read(p []byte) (n int, err error) {
 		} else {
 			bufferLog.Debugf("read %v from buffer", n)
 		}
-	} else if b.err != nil {
-		err = b.err
 	} else {
 		n, err = b.readFromStream(p)
 	}
--- a/internal/buffer_pool_test.go
+++ b/internal/buffer_pool_test.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"io"
 	"sync"
+	"syscall"
 	"time"
 
 	"github.com/sirupsen/logrus"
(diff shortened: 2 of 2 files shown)


kahing:
I fixed it in a different way but thanks for the bug report!

inevity:
Once update to the lasted commit on master, goofys cannot pass the testcase i have supplied in this pull request. It still panic on same code line.

Also i compose a testcase:
which read 45MB file .this will have two readbuffer ,first range is 20-40MB,second range is 40-45MB.
For the second readahead,the start offset will be 2* 20971520=40MB, i fake a permission denid by chang the s3 req time smaller than sys time in buffer init function.
The test case will read file by 128KB io block. 
The test case will alse panic .

271 func (b S3ReadBuffer) Init(fh *FileHandle, offset uint64, size uint32) *S3ReadBuffer {
if offset == uint64(20971520) {    
//if change the 20MB chunk to 10MB to quick test,so the offset is  20971520                                                                                                                                                                        
301       req.Time = req.Time.AddDate(-1, 2, 3)
305     }
308     err := req.Send()



note:
proxys3 does not have the time diff  function,so it will not test the issuse. I use the minio backend and another storage alike aws s3.  
@kahing

inevity:
when cutover to parrell readahead, the second readahead buffer's init function will first do s3 op,whill will lead err =13(permisson denied). Later,when the io have reach the second readbuf offset, it will run buffer.read. 
At this time ,b.reader == nil and b.err == 13. buf!=nil.
the for continion "for b.reader == nil && b.err == nil "will skip ,the code will also go the  b.readFromStream .then will panic.

inevity:
maybe we change here
func (b *Buffer) readLoop(r ReaderProvider) {
	for {
		b.mu.Lock()
		if b.reader == nil {
			b.reader, b.err = r()
			b.cond.Broadcast()
			if b.err != nil {
				b.mu.Unlock()
                                b.buf.Close() //add this 
				break
			}

gaul:
> proxys3 does not have the time diff function,so it will not test the issuse. I use the minio backend and another storage alike aws s3.

Can you expand on this comment?  I don't understand what S3Proxy lacks.

inevity:
@gaul 
In aws s3,
there is a timestamp requirments. It say "Time Stamp Requirement

A valid time stamp (using either the HTTP Date header or an x-amz-date alternative) is mandatory for authenticated requests. Furthermore, the client timestamp included with an authenticated request must be within 15 minutes of the Amazon S3 system time when the request is received. If not, the request will fail with the RequestTimeTooSkewed error code. The intention of these restrictions is to limit the possibility that intercepted requests could be replayed by an adversary. For stronger protection against eavesdropping, use the HTTPS transport for authenticated requests." 
ref :http://docs.aws.amazon.com/AmazonS3/latest/dev/RESTAuthentication.html#RESTAuthenticationTimeStamp.

If use the testcase about fake the time to send req i describe above,proxys3 didnot return permission denied,but minio and my aws s3 like storage does. so i think proxys3 lack the function to validate the time diff between the req time with aws sys time.

kahing:
thanks for the report @inevity !

gaul:
@inevity Could you report an issue against S3Proxy?  It should handle this time stamp skew and there are tests for this but perhaps it doesn't check one of the three sources of time that you use.

PR #255, opened 2017-12-28
do not use STANDARD_IA storage class for files less than 128KB

jtwang83:
STANDARD_IA storage class has minimum charge of 128KB, we should not use STANDARD_IA unless the object size is greater than 128KB. This also increases cost efficiency for operations that first open an empty file and then write to it (like shell redirection).


kahing:
amended commit with fix suggested by @gaul . Also please run the code through go-fmt in the future.

PR #292, opened 2018-03-21
Add support for request-payer buckets

msmitherdc:
Added a new parameter for a bucket, boolean, which allows a user to connect to a requester-payer bucket. Also updated the aws go sdk version due to some bugs with request payer support in the api.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -38,6 +38,7 @@ type Config struct {
 	Region         string
 	RegionSet      bool
 	StorageClass   string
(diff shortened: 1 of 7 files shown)


kahing:
Sorry for the delay in looking into this. Does it make more sense to inject a handler to set the requester pay param (see `addAcceptEncoding` in goofys.go). I don't doubt that you've added it to all the necessary places but I worry that future code changes will break this.

PR #297, opened 2018-03-30
Add a parameter max-keys to tuning

zhezhel:
Added a new tuning parameter, max-keys (int64), for speed up ls in big buckets.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -46,8 +46,10 @@ type Config struct {
 	ACL            string
 
 	// Tuning
-	Cheap        bool
-	ExplicitDir  bool
+	Cheap       bool
+	ExplicitDir bool
+	MaxKeys     int64
+
 	StatCacheTTL time.Duration
 	TypeCacheTTL time.Duration
 
--- a/internal/dir.go
+++ b/internal/dir.go
@@ -244,6 +244,7 @@ func (dh *DirHandle) listObjects(prefix string) (resp *s3.ListObjectsOutput, err
 			Bucket:    &fs.bucket,
 			Delimiter: aws.String("/"),
 			Marker:    dh.Marker,
+			MaxKeys:   aws.Int64((*fs.flags).MaxKeys),
 			Prefix:    &prefix,
 		}
 
--- a/internal/flags.go
+++ b/internal/flags.go
@@ -204,6 +204,11 @@ func NewApp() (app *cli.App) {
 				Name:  "no-implicit-dir",
 				Usage: "Assume all directory objects (\"dir/\") exist (default: off)",
 			},
+			cli.Int64Flag{
+				Name:  "max-keys",
+				Value: int64(1000),
+				Usage: "Maximum number of keys returned per one connection in the response body",
+			},
 
 			cli.DurationFlag{
 				Name:  "stat-cache-ttl",
@@ -250,7 +255,7 @@ func NewApp() (app *cli.App) {
 		flagCategories[f] = "aws"
 	}
 
-	for _, f := range []string{"cheap", "no-implicit-dir", "stat-cache-ttl", "type-cache-ttl"} {
+	for _, f := range []string{"cheap", "no-implicit-dir", "max-keys", "stat-cache-ttl", "type-cache-ttl"} {
 		flagCategories[f] = "tuning"
 	}
 
@@ -295,6 +300,7 @@ type FlagStorage struct {
 	// Tuning
 	Cheap        bool
 	ExplicitDir  bool
+	MaxKeys      int64
 	StatCacheTTL time.Duration
 	TypeCacheTTL time.Duration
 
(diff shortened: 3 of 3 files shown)


zhezhel:
Test failed on the line github.com/kahing/goofys/internal/goofys.go:726
	// XXX/is this a dir?
	dh := in.OpenDir()

kahing:
Why do you want this? I believe the default is already 1000 which is S3 max.

pawelmarkowski:
@kahing first part of your assumption is correct. Default is 1000, but max-keys is supported for instance by latest s3cmd (2.0.1). It is important parameter in S3 standard.

kahing:
What's the intended use case to use a different Max key?

pawelmarkowski:
To speedup listing of big amount objects. https://docs.aws.amazon.com/AmazonS3/latest/API/v2-RESTBucketGET.html it is part of the standard. We have an thousands of objects in our storage - splitting it into parts per 1000 objects works slowly.

kahing:
That's v2 but regardless: "This implementation of the GET operation returns some or all (up to 1,000) of the objects in a bucket."

pawelmarkowski:
https://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketGET.html  ... v1 the same . It is part of the standard. And I do not understand where is a problem to merge this part of code. Could you explain that? As I told you if you have over 1 000 objects than response is split into parts ... so if I open goofys mount and I have a space with 40 000 objects - goofys asks about this objects 40 times. It is not most efficient solution at all. With this patch I am able to set max keys for instance to 20 000, and I will receive response split into 2 parts.

kahing:
According to the documentation that I pasted the maximum number of objects returned is 1000, so using a max-key of 20000 should have no effect. Are you claiming otherwise?

kahing:
Also, are you using a non-AWS s3 implementation? If so which one?

pawelmarkowski:
@kahing read the docs one more time, or just test goofys on path, where you have over 1000 objects. I have described you use case - on each of my previous comments. Nevertheless, once again: I have local object storage that is compatible with s3 implementations, which contains thousands of objects. Please merge this request, because I have checked that change and in my environment it works perfectly. ls operatoion time has decreased from 14s to 6.8s.

Max-keys parameter is available in latest s3cmd because is part of standard. It was available in previous versions too (s3cmd --help), but was not implement in a code. Right now it is. 

I appreciate your effort to investigate usefulness of this parameter. It is really crucial for object storage collections that contains thousands or even millions of objects. Without this tuning parameter - goofys efficiency will be not enough.

kahing:
@pawelmarkowski Please stop telling me to read the documentation. I do want to help you to use goofys in an efficient manner. Which local object storage are you using, and what's the highest max keys that it supports? Obviously I am not able to test this myself because as I stated, AWS S3 does NOT support max keys higher than 1000.

One reason why I am hesitate to add an option is because I want goofys to be fast for as many people as possible without additional configuration. For example, would it be possible to detect the object storage that you use and use a higher max keys? Or simply always use a higher max keys? aws-s3 for example would just treat larger max keys as the same as 1000. So maybe that's something goofys can do by default without having a new option.

pawelmarkowski:
@kahing I was slightly annoyed, because I thought we are spinning in circle. I am sorry. Nevertheless we have an Object Storage (Ceph) with over 2 billions objects. Some of listings (ls /some/example/path/) should return over 50 thousands objects. Splitting the S3 query to the 1 000 parts is not such efficient as we want.

In my opinion it would be a good solution (practice from other tools s3cmd and so on) to add this parameter (--max-keys)  as an optional for advanced users. Default value should stay 1000. If user do not use it, it stays transparent. I am not sure are we able to predict --max-keys value to ensure best efficiency. Storage may be changed in couple days and this parameter should be managed by system administrator.

kahing:
Ok lets do 3 things:

* add a new help section (similar to the one for "AWS S3 OPTIONS") for non-AWS s3 to make it clear that it's only useful for alternative S3 implementations. "NON-AWS S3 OPTIONS" doesn't roll off the tongue but I am okay with that if no one can come up with anything better

* default max-keys to 10K since aws doesn't mind us passing a larger value anyway (bonus point if you want to detect ceph automatically and default to 10K only for ceph)

* make the tests pass (let me know if you need more help with this)

kahing:
My reason for resisting to "simply" add an option for max-keys is because one of goofys' motivation is "fast by default" instead of having to read a lot of documentations to understand all the options. Cloud/object uses cases are increasingly enabling self-serve dev-op style work flows where storage admins are removed from day to day developments. Many of goofys users don't understand S3 as well as the most advanced users and many of them use S3 precisely because they don't have a system admin.

I understand your frustration that I am asking you to do more work even though the extra work doesn't benefit you, but I am just trying to make goofys more useful for people who are not as resourceful as you guys.

pawelmarkowski:
@kahing thanks for reply. It is a good idea. I think that the value of max-keys could be the same even for NON-AWS S3 (1 000 is OK) ... when an user needs to modify this value then it can use max-keys parameter.

gaul:
I would request 1,000 by default since S3 clones may not allow higher values.

PR #303, opened 2018-04-07
Support Minio SSE-C

MikeHawkesCapventis:
We would like to employ goofys on top of Minio but due to incoming EU data protection laws would like to encrypt by default. Minio supports SSE-C, please will you add this to the options already available?


kahing:
is there a reason why this was based on the append branch and not master?

kahing:
hmm this doesn't seem to be an actual PR, closing

PR #314, opened 2018-04-20
goofys when run in Docker/POD uses the HOST memory limits

eswarantg:
The memory allocation logic within goofys uses HOST memory limits from the /proc/meminfo file on Linux.  This is actually the Host parameter.

In a POD word the POD memory limit via cgroup information setup.
/sys/fs/cgroup/memory{containerID}/memory.limit_in_bytes

The containerId is acquired from 
/proc/self/cpuset

But to determine if the goofys is being run in a Host based env or Docker based env
/proc/1/sched file is checked for determining if the process is init or something else.

### DIFF
--- a/internal/buffer_pool.go
+++ b/internal/buffer_pool.go
@@ -15,15 +15,15 @@
 package internal
 
 import (
+	"errors"
 	"io"
+	"io/ioutil"
+	"path/filepath"
 	"runtime"
 	"runtime/debug"
-	"sync"
-	"io/ioutil"
-	"strings"
 	"strconv"
-	"path/filepath"
-    "errors"
+	"strings"
+	"sync"
 
 	"github.com/jacobsa/fuse"
 	"github.com/shirou/gopsutil/mem"
@@ -50,12 +50,12 @@ func maxMemToUse(buffersNow uint64) uint64 {
 		panic(err)
 	}
 
-    availableMem, err := getCgroupAvailableMem()
+	availableMem, err := getCgroupAvailableMem()
 	log.Debugf("amount of available memory from cgroup is: %v", availableMem/1024/1024)
 
-    if(err != nil || availableMem < 0 || availableMem > m.Available) {
-        availableMem = m.Available
-    }
+	if err != nil || availableMem < 0 || availableMem > m.Available {
+		availableMem = m.Available
+	}
 
 	log.Debugf("amount of available memory: %v", availableMem/1024/1024)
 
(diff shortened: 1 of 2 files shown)


internal/DockerMemoryLimit.go
```diff
@@ -0,0 +1,188 @@
...
+	d._memLimitInBytesFile = "/memory.limit_in_bytes"
+	return
+}
+
+func (d *DockerMemoryLimit) init() {
+	if !d.readDone {
+		d.readlock.Lock()
+		defer d.readlock.Unlock()
+		if !d.readDone {
+			d.readIsDocker()
```
kahing: it shouldn't matter that we are in docker, we should use memory cgroup limit if there is one

internal/DockerMemoryLimit.go
```diff
@@ -0,0 +1,188 @@
...
+		d.memLimitInBytesError = errors.New(filename + "\"" + line + "\" IntConv failed:%s" + err.Error())
+		log.Error(d.memLimitInBytesError)
+		return
+	}
+	d.memLimitInBytes = uint64(limit)
+	d.memLimitInBytesError = nil
+	log.Infof("%s memLimitInBytes=%d", filename, d.memLimitInBytes)
+}
+func (d *DockerMemoryLimit) IsDocker() bool {
+	d.init()
```
kahing: init() once instead of calling init() everywhere

internal/file.go
```diff
@@ -164,7 +164,7 @@ func (fh *FileHandle) mpuPartNoSpawn(buf *MBuf, part int, total int64, last bool
 			Body:       buf,
 		}
 
-		s3Log.Debug(params)
+		s3Log.Info(params)
```
kahing: remove these changes

internal/buffer_pool.go
```diff
@@ -40,19 +41,41 @@ type BufferPool struct {
 const BUF_SIZE = 5 * 1024 * 1024
 
 func maxMemToUse(buffersNow uint64) uint64 {
-	m, err := mem.VirtualMemory()
-	if err != nil {
-		panic(err)
+	var max uint64
+	var containerMemLimitUsed bool
+	if IsDocker() {
+		limit, err := GetMemLimitInBytes()
```
kahing: is it possible for the limit to change at run time?

internal/buffer_pool.go
```diff
@@ -40,19 +41,41 @@ type BufferPool struct {
...
+		if err != nil {
+			log.Errorf("Error getting GetMemLimitInBytes:%s", err)
+		} else {
+			if limit != math.MaxUint64 {
+				containerMemLimitUsed = true
+				log.Debugf("amount of total memory: %v", limit/1024/1024)
```
kahing: refactor to remove the duplication here

kahing:
thanks for the contribution! I looked over it and have a few suggestions.

yuvisara:
Hi @kahing, can you please review  these changes and give an official build if possible?

kahing:
could you squash the commits? it's really hard to review the diff when you've pretty much rewritten everything

yuvisara:
@kahing squashed both the commits.

eswarantg:
@kahing Can you please review the changes and approve?

internal/buffer_pool.go
```diff
@@ -45,19 +50,107 @@ func maxMemToUse(buffersNow uint64) uint64 {
 		panic(err)
 	}
 
-	log.Debugf("amount of available memory: %v", m.Available/1024/1024)
+    availableMem := getCgroupAvailableMem()
+	log.Debugf("amount of available memory from cgroup is: %v", availableMem/1024/1024)
```
WD-stefaang: tabs vs spaces

eliastor:
How long it'll take to merge this pull-request? It's reallly useful

internal/buffer_pool.go
```diff
@@ -45,19 +50,107 @@ func maxMemToUse(buffersNow uint64) uint64 {
...
+        return
+    }
+
+    mem_usage, err := readFileAndGetValue(filepath.Join(CGROUP_FOLDER_PREFIX, path, MEM_USAGE_FILE_SUFFIX))
+    if err != nil {
+        log.Debugf("Unable to get memory usage from cgroup error: %v", err)
+        return
+    }
+
+    return (mem_limit - mem_usage)
```
kahing: mem_usage is  `memory.max_usage_in_bytes	 # show max memory usage recorded` according to https://www.kernel.org/doc/Documentation/cgroup-v1/memory.txt , which sounds like it will show the peak usage even if the current usage is less? seems to me that you want to get the memory limit from cgroup and check the current usage the old fashion way

kahing:
> How long it'll take to merge this pull-request? It's reallly useful

sorry been busy and overlooked this, I added a few comments

internal/buffer_pool.go
kahing: yes usage_in_bytes sounds more correct than max_usage_in_bytes. I don't mind that we pull in an external dependency for this

internal/buffer_pool.go
```diff
@@ -45,19 +50,107 @@ func maxMemToUse(buffersNow uint64) uint64 {
...
-	max := uint64(m.Available+ms.Sys) / 2
+	max := uint64(availableMem+ms.Sys) / 2
 	maxbuffers := MaxUInt64(max/BUF_SIZE, 1)
 	log.Debugf("using up to %v %vMB buffers, now is %v", maxbuffers, BUF_SIZE/1024/1024, buffersNow)
 	return maxbuffers
 }
 
+const CGROUP_PATH = "/proc/self/cgroup"
+const CGROUP_FOLDER_PREFIX = "/sys/fs/cgroup/memory" 
+const MEM_LIMIT_FILE_SUFFIX = "/memory.limit_in_bytes"
```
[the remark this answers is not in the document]
yuvisara: @kahing do you mean add the following link in the comments?
https://www.kernel.org/doc/Documentation/cgroup-v1/memory.txt

yuvisara:
@kahing Have addressed the review comments. Will squash the commits if it is required.

eswarantg:
@kahing i see the requests are resolved by @yuvisara.  Can we get this closed and merged?
Also, I see last release is Jan 2018.  Is there a plan to create another release?

kahing:
@eswarantg definitely. Give me a few days. Started a new job a month ago and it's related to goofys, so activity should hopefully go up again.

kahing:
so I looked into this some more, since at $dayjob we run goofys in a LXC container with memory cgroup limits. A couple findings:

1. we don't have access to `/sys/fs/cgroup/memory` from inside the container, as the "modern" way is to access via cgmanager: https://lwn.net/Articles/618411/
1. for us `/proc/meminfo` is actually correct, because we use https://github.com/lxc/lxcfs takes care of it.

In docker though, this seems to be the right thing to do (docker mounts `cgroupfs` into the container and doesn't use `lxcfs` for `/proc/meminfo`).

kahing:
merged via c927f2e188a0f5f5870afe83bb161dcd077bafa1

yuvisara:
👍

PR #330, opened 2018-06-20
Expose NewGoofys in `github.com/kahing/goofys/api` package

bgardiner:
I'm currently extending `goofys` to add local caching for some specific use cases in which `catfs` does not serve our needs.  However, the current API of the `goofys/api` package is very limited (as it only allows you to mount).  It would be really useful to be able to create a  `goofys/internal.Goofys` struct such that certain functions of the `github.com/jacobsa/fuse/fuseutil.FileSystem` interface can be overridden; but the associated goofys functions and types are currently walled off in the `goofys/internal` package, which cannot be imported.

This PR exposes a few of the functions and types from the `internal` package that enable extending `goofys`. I also added a `GetFullName` function I've found very useful.  This PR makes no functional changes but would be awesome to have incorporated into the codebase from an extensibility standpoint.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -1,7 +1,7 @@
 package goofys
 
 import (
-	. "github.com/kahing/goofys/internal"
+	"github.com/kahing/goofys/internal"
 
 	"context"
 	"fmt"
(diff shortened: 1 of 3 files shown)


bensussman:
Any update on this?

api/api.go
```diff
@@ -178,3 +178,21 @@ func Mount(
 
 	return
 }
+
+// expose Goofys related functions and types for extending and mounting elsewhere
+var (
+	GetStdLogger       = internal.GetStdLogger
+	InitLoggers        = internal.InitLoggers
+	MassaageMountFlags = internal.MassageMountFlags
```
kahing: misspelled massage

bgardiner:
ping @kahing -- I think it's ready for merge

PR #334, opened 2018-06-28
Set a timeout on the http client

### LINKED ISSUE #329: Stuck IO operations
I'm trying to use `goofys` to mount an OpenStack container. I don't know any details of the setup. Read performance is generally good, but from time to time processes get stuck while reading some files.

This shows up under `strace` as a `read` that never returns. I can reproduce it with sequential workloads like `cat` or `md5sum`, which get stuck in the middle of the file, at the same position. There might be some caching involved (but not `catfs`). I didn't try to `strace` the `goofys` process, but I don't think there are any errors in the logs.

Accessing the same files from another machine works fine. After unmounting and remounting the container, they start working again. I have reason to believe that the server may be unstable, for some definition of it.

I didn't look into the implementation of `goofys`. Maybe this could be fixed by setting a timeout on the HTTP client? I would be fine with `EIO` or something like that in this case.

lnicola:
Fixes #329

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -83,6 +83,7 @@ func Mount(
 			TLSHandshakeTimeout:   10 * time.Second,
 			ExpectContinueTimeout: 10 * time.Second,
 		},
+        Timeout: 30 * time.Second,
 	})
 
 	if len(flags.Profile) > 0 {


lnicola:
https://github.com/kahing/goofys/commit/f7834cf8cb786f7c978b01ff281ba5fcd887ec20

PR #351, opened 2018-08-22
add subdomain and ssl option

zhenlingcn:
add subdomain and ssl option

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -44,6 +44,8 @@ type Config struct {
 	UseKMS         bool
 	KMSKeyID       string
 	ACL            string
+	Subdomain      bool
+	UseSSL         bool
 
 	// Tuning
 	Cheap        bool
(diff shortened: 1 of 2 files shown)


kahing:
This would disable SSL by default which is definitely not wanted. You can already disable SSL by passing a http `--endpoint`.

What's the motivation for disabling path based access?

zhenlingcn:
some object storage doesn't support path based access,we need provide an option.

zhenlingcn:
I had removed ssl option,but I think the subdomain option is necessary,please merge it,thanks.

kahing:
This is fine, could you squash and also change the option and description to be more clear? The "subdomain" mode is called virtual-hosted style: https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingBucket.html#access-bucket-intro . s3fs has `-o use_path_request_style` for the inverse behavior. I suggest using `--use-vhost` instead of `--subdomain`.

kahing:
I must have been half-asleep when I merged this as 2883047a86f659b77551e384156a4ed09091393e. I still like `--use-vhost` more which #403 added.

PR #356, opened 2018-09-12
Add support for Requester Pays buckets

lnicola:
Closes #292.

Includes a drive-by fix by @msmitherdc (I assume).

r? @msmitherdc.

#292 also has a test for the stat cache, do we want to include it?

### DIFF
--- a/internal/goofys.go
+++ b/internal/goofys.go
@@ -236,8 +236,19 @@ func addAcceptEncoding(req *request.Request) {
 	}
 }
 
+func addRequestPayer(req *request.Request) {
+	// "Requester Pays" is only applicable to these
+	// see https://docs.aws.amazon.com/AmazonS3/latest/dev/RequesterPaysBuckets.html
+	if req.HTTPRequest.Method == "GET" || req.HTTPRequest.Method == "HEAD" || req.HTTPRequest.Method == "POST" {
+		req.HTTPRequest.Header.Set("x-amz-request-payer", "requester")
+	}
+}
+
 func (fs *Goofys) newS3() *s3.S3 {
 	svc := s3.New(fs.sess)
+	if fs.flags != nil && fs.flags.RequesterPays {
+		svc.Handlers.Build.PushBack(addRequestPayer)
+	}
 	if fs.v2Signer {
 		svc.Handlers.Sign.Clear()
 		svc.Handlers.Sign.PushBack(SignV2)
--- a/api/api.go
+++ b/api/api.go
@@ -38,6 +38,7 @@ type Config struct {
 	Region         string
 	RegionSet      bool
 	StorageClass   string
+	RequesterPays  bool
 	Profile        string
 	UseContentType bool
 	UseSSE         bool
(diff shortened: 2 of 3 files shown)


lnicola:
Some help with the test suite, please? I don't know what any of that means.

On my machine I get:

```
$ test/run-tests.sh
# github.com/kahing/goofys/internal
internal/goofys_test.go:45:2: cannot find package "github.com/kahing/go-xattr" in any of:
	/home/grayshade/go/src/github.com/kahing/goofys/vendor/github.com/kahing/go-xattr (vendor tree)
	/usr/lib/go/src/github.com/kahing/go-xattr (from $GOROOT)
	/home/grayshade/go/src/github.com/kahing/go-xattr (from $GOPATH)
FAIL	github.com/kahing/goofys/internal [setup failed]
?   	github.com/kahing/goofys	[no test files]
?   	github.com/kahing/goofys/api	[no test files]
?   	github.com/kahing/goofys/example	[no test files]
```

lnicola:
@kahing 

Tests are now passing. Does this look all right?

internal/handles.go
```diff
@@ -200,8 +200,7 @@ func (parent *Inode) removeChildUnlocked(inode *Inode) {
 	}
 	i := sort.Search(l, parent.findInodeFunc(*inode.Name, inode.isDir()))
 	if i >= l || *parent.dir.Children[i].Name != *inode.Name {
-		panic(fmt.Sprintf("%v.removeName(%v) but child not found: %v",
```
kahing: probably shouldn't be part of this?

internal/goofys.go
```diff
@@ -236,8 +236,17 @@ func addAcceptEncoding(req *request.Request) {
 	}
 }
 
+func addRequestPayer(req *request.Request) {
+	if req.HTTPRequest.Method == "GET" || req.HTTPRequest.Method == "HEAD" || req.HTTPRequest.Method == "POST" {
```
kahing: please add a comment on why only these 3 methods

internal/handles.go
lnicola: It came from https://github.com/kahing/goofys/pull/292. I'll remove it.

internal/goofys.go
kahing: I meant we should document (if any) methods goofys use that are not requester pays.

Also, from that link seems like we should disable anonymous bucket access as well

lnicola: I'm not sure what you mean. I never used S2 (sorry if you're aware what I'll be writing), but I believe requester pays is mostly used for buckets storing large datasets, which the owner wants the data to be public, but doesn't want to pay for the transfer costs. The header is an acknowledgement that the user (caller) knows they're paying for the transfer, and that's why it's (only) required for HTTP methods that retrieve data.

Without the header (and some credentials), the bucket won't be accessible at all. We _could_ raise an error on startup if the user tried to enable the requester pays mode without providing credentials, I assume.

kahing: What I meant is, why are you checking the request methods are GET/HEAD/POST?

kahing:
@lnicola sorry for the very delayed merge and thanks for the PR!

lnicola:
No worries. I was actually planning to revisit this soon.

PR #360, opened 2018-09-20
Add command line option to override http.Client Timeout

pcram-techcyte:
We use goofys to do random reads in large files stored in S3.  Most of the time, the GetObject requests to S3 are very fast - less than 100 ms.  But occasionally - about 1 in 1000 requests - it takes much longer.  Several seconds, even up to 20 seconds.  The 30 second timeout is nice, but in our case we've found that a much shorter timeout and quick retry works better for us - as low as 1 second really helped even out our response times.

I've added a command line flag `-http-timeout` to set the value, but it has the same default as the previous hard-coded value (30 seconds).  So it should behave exactly the same if you don't specify the flag.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -50,6 +50,7 @@ type Config struct {
 	ExplicitDir  bool
 	StatCacheTTL time.Duration
 	TypeCacheTTL time.Duration
+	HTTPTimeout  time.Duration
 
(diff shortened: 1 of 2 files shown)


kahing:
thanks for the contribution @pcram-techcyte

PR #361, opened 2018-09-23
Allow static credentials in API

djmaze:
This is needed so the [Docker plugin](https://github.com/djmaze/goofys-docker) can run flawlessly in a Docker swarm cluster.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -38,6 +38,8 @@ type Config struct {
 	Region         string
 	RegionSet      bool
 	StorageClass   string
+	AccessKey      string
+	SecretKey      string
 	Profile        string
 	UseContentType bool
 	UseSSE         bool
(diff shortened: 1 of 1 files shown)


kahing:
sorry for the late response. Could you briefly explain why you cannot use static credentials through the ~/.aws/credentials file in this case?

djmaze:
Running in a swarm cluster means potentially using different credentials for different buckets. Docker volumes are configured individually on creation, so the plugin would need to manage a credentials file on each server (using locking in order to prevent an inconsistent state).

So while it is certainly possible, the API-based solution is much cleaner.

kahing:
but you can use one credentials file with different profiles, right?

djmaze:
Yes, but as said that would be cumbersome. 

I get that the credentials file makes sense for commandline usage. But an API shouldn't need a separate file to be managed for configuration, in my opinion.

djmaze:
Thanks!

PR #373, opened 2018-10-26
add credentials in flags

flrossetto:


gaul:
This is insecure since the credentials flags show up in process listing.  Why doesn't the `${HOME}/.aws/credentials` file suffice?  And if it does not, would environment variable address your use case better?

flrossetto:
`$ {HOME} /. Aws / credentials` is not sufficient and env does not work.

nigels-com:
The env variables seem to be working for me.

I agree with not using credentials on the CLI in production, but not so opposed to supporting it along with a suitably stern warning message.

flrossetto:
modified to get env value.

kahing:
I don't understand this change, what problems are you trying to solve?

Issue #392, opened 2019-01-07
ForgetInode race condition lead to "panic: Unknown inode"?


zhjunqin:
I met following issue when use goofys.

```
2019/01/07 20:55:18.486805 fuse.DEBUG Op 0x0002b709        connection.go:395] <- LookUpInode (parent 4, name "record-00066-of-00100")
2019/01/07 20:55:18.487230 fuse.DEBUG Inode.LookUp 4 record/files [record-00066-of-00100]
2019/01/07 20:55:18.493826 fuse.DEBUG <-- LookUpInode 4 coco_train.record-00066-of-00100 <nil>
2019/01/07 20:55:18.494513 fuse.DEBUG Op 0x0002b709        connection.go:478] -> OK (inode 1985)
2019/01/07 20:55:18.494815 fuse.DEBUG Op 0x0002b70a        connection.go:395] <- LookUpInode (parent 4, name "record-00066-of-00100")
2019/01/07 20:55:18.494916 fuse.DEBUG Op 0x0002b70b        connection.go:395] <- ForgetInode (inode 1985)
2019/01/07 20:55:18.494999 fuse.DEBUG DeRef 1985 record/files/record-00066-of-00100 [1 1]
2019/01/07 20:55:18.495071 fuse.DEBUG Ref 1985 record/files/record-00066-of-00100 [0]
2019/01/07 20:55:18.495098 fuse.DEBUG <-- LookUpInode 4 record-00066-of-00100 <nil>
2019/01/07 20:55:18.495134 fuse.DEBUG Op 0x0002b70a        connection.go:478] -> OK (inode 1985)
2019/01/07 20:55:18.495230 fuse.DEBUG Op 0x0002b70b        connection.go:478] -> OK ()
2019/01/07 20:55:18.495271 fuse.DEBUG Op 0x0002b70c        connection.go:395] <- OpenFile (inode 1985)
panic: Unknown inode: 1985

goroutine 243042 [running]:
github.com/kahing/goofys/internal.(*Goofys).getInodeOrDie(0xc420242340, 0x7c1, 0xc4204dbf60)
        /opt/bin/goofys/work/src/github.com/kahing/goofys/internal/goofys.go:421 +0x126
github.com/kahing/goofys/internal.(*Goofys).OpenFile(0xc420242340, 0xa01b20, 0xc429697a70, 0xc42915c860, 0x0, 0x0)
[...]

PR #394, opened 2019-01-08
Fix race condition between LookUpInode and ForgetInode

zhjunqin:
This pull request try to fix issue https://github.com/kahing/goofys/issues/392

```
1. ForgetInode get fs.mu.Lock()
2. LookUpInode get parent.mu.Lock()
3. ForgetInode release fs.mu.Unlock() and wait for parent.mu.Lock() in inode.Parent.removeChild(inode)
4. LookUpInode get fs.mu.Lock() then get inode
5. LookUpInode release parent.mu.Unlock()
6. inode.Parent.removeChild(inode) get parent.mu.Lock() and remove the inode
```

If I just change `ForgetInode`，which will lead to other deadlock, that is 
```
1. ForgetInode get fs.mu.Lock()
2. LookUpInode get parent.mu.Lock()
3. ForgetInode try to get parent.mu.Lock()
4. LookUpInode get fs.mu.Lock()
```

So I think `LookUpInode` also needs change order of lock. But I'm not familiar with goofys, please help review it. 

Thank you very much!

### DIFF
--- a/internal/goofys.go
+++ b/internal/goofys.go
@@ -604,8 +604,8 @@ func (fs *Goofys) LookUpInode(
 	parent := fs.getInodeOrDie(op.Parent)
 	fs.mu.Unlock()
 
-	parent.mu.Lock()
 	fs.mu.Lock()
+	parent.mu.Lock()
 	inode = parent.findChildUnlockedFull(op.Name)
 	if inode != nil {
 		ok = true
@@ -627,8 +627,8 @@ func (fs *Goofys) LookUpInode(
 	} else {
 		ok = false
 	}
-	fs.mu.Unlock()
 	parent.mu.Unlock()
+	fs.mu.Unlock()
 
 	if !ok {
 		var newInode *Inode
@@ -699,14 +699,12 @@ func (fs *Goofys) ForgetInode(
 	if stale {
 		delete(fs.inodes, op.Inode)
 		fs.forgotCnt += 1
-		fs.mu.Unlock()
 
 		if inode.Parent != nil {
 			inode.Parent.removeChild(inode)
 		}
-	} else {
-		fs.mu.Unlock()
 	}
+	fs.mu.Unlock()
 
 	return
 }
--- a/internal/goofys.go
+++ b/internal/goofys.go
@@ -650,17 +650,17 @@ func (fs *Goofys) LookUpInode(
 		}
 
 		if inode == nil {
+			fs.mu.Lock()
 			parent.mu.Lock()
 			// check again if it's there, could have been
 			// added by another lookup or readdir
 			inode = parent.findChildUnlockedFull(op.Name)
 			if inode == nil {
-				fs.mu.Lock()
 				inode = newInode
 				fs.insertInode(parent, inode)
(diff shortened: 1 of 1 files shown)


kahing:
Thanks for the PR! Could you describe how I could reproduce this (and ideally a test case)?

Admittedly lock ordering is not well thought out. My intention was that for locking `A/B`, lock `B` before locking `A/` before locking global lock (if necessary). Your PR changes the lock ordering to lock `fs.mu` before `parent.mu` which I worry about - inconsistent lock ordering can lead to deadlock. Do you think reordering the lock in your PR could work?

zhjunqin:
> Thanks for the PR! Could you describe how I could reproduce this (and ideally a test case)?
> 
> Admittedly lock ordering is not well thought out. My intention was that for locking `A/B`, lock `B` before locking `A/` before locking global lock (if necessary). Your PR changes the lock ordering to lock `fs.mu` before `parent.mu` which I worry about - inconsistent lock ordering can lead to deadlock. Do you think reordering the lock in your PR could work?

Thanks for reply.

The problem happened randomly, so there is no a determine test case to reproduce it.
We run goofys about more than 5 hours, then the problem would probably happen.

As I said, the root cause is following race condition
```
1. ForgetInode get fs.mu.Lock()
2. LookUpInode get parent.mu.Lock(), wait for fs.mu.Lock()
3. ForgetInode release fs.mu.Unlock() and wait for parent.mu.Lock() in inode.Parent.removeChild(inode)
4. LookUpInode get fs.mu.Lock() then get inode
5. LookUpInode release parent.mu.Unlock() return the Inode
6. inode.Parent.removeChild(inode) get parent.mu.Lock() and then remove the inode
7. then lead to "panic: Unknown inode" problem
```
I think the problem can be triggered by simulating the above race condition.
But I don't know how to trigger the ForgetInode.

zhjunqin:
If there is any other solution to solve the race condition problem, I would be glad to try and do test.

internal/goofys.go
```diff
@@ -604,8 +604,8 @@ func (fs *Goofys) LookUpInode(
 	parent := fs.getInodeOrDie(op.Parent)
 	fs.mu.Unlock()
 
-	parent.mu.Lock()
```
kahing: is there a reason why this order is changed?

zhjunqin: In function `ForgetInode` lock order is `fs.mu.Lock()` -> `parent.mu.Lock()`
In function `LookUpInode`, if still keep old order  `parent.mu.Lock()` -> `fs.mu.Lock()`, would lead to deadlock.

That is function `ForgetInode` get `fs.mu.Lock()` and wait for  `parent.mu.Lock()`, at same time function `LookUpInode` get `parent.mu.Lock()` and wait for `fs.mu.Lock()`.

zhjunqin: Same reason for code change in another commit.

I find the deadlock problem when running with only code change in function `ForgetInode`.

kahing:
@zhjunqin and @popsUlfr could you see if the commit in #410 solves the problems you see?

PR #404, opened 2019-03-11
update ci settings: base distribution and java/golang version

t2y:
Current travis settings is too old, so I updated.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,10 @@
 language: go
 sudo: required
-dist: trusty
-addons:
-  apt_packages:
-   - openjdk-7-jre-headless
+dist: xenial
 env:
-- GO15VENDOREXPERIMENT=1
 install:
 - go get -t $(go list ./... | grep -v /vendor/)
 - mkdir /tmp/mnt
 - make
 go:
-  - 1.10.3
+  - 1.11.5
--- a/test/run-tests.sh
+++ b/test/run-tests.sh
@@ -20,6 +20,7 @@ trap cleanup EXIT
 rm -Rf /tmp/s3proxy
 mkdir -p /tmp/s3proxy
 
+java -version
 export LOG_LEVEL=warn
 PROXY_BIN="java -jar s3proxy.jar --properties test/s3proxy.properties"
 stdbuf -oL -eL $PROXY_BIN &
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ run-test: s3proxy.jar
 	./test/run-tests.sh
 
 s3proxy.jar:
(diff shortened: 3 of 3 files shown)


t2y:
I tried to use java8 with below settings instead of java11 (by default) for conservative migration. But, some configurations are not useful to change java version.

```
before_install:
  - sudo apt-get update
  - sudo apt-get install -y openjdk-8-jre-headless
  - sudo update-java-alternatives --jre-headless --jre --set java-1.8.0-openjdk-amd64
```

https://github.com/travis-ci/travis-ci/issues?utf8=%E2%9C%93&q=is%3Aissue+openjdk-8+xenial

I don't have ubuntu environment, so I cannot reproduce it. However, s3proxy.jar also works with java11 though it shows some warnings. IIRC, this behavior is changed since java9.

```
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by com.google.inject.internal.cglib.core.$ReflectUtils$2 (file:/home/travis/gopath/src/github.com/kahing/goofys/s3proxy.jar) to method java.lang.ClassLoader.defineClass(java.lang.String,byte[],int,int,java.security.ProtectionDomain)
WARNING: Please consider reporting this to the maintainers of com.google.inject.internal.cglib.core.$ReflectUtils$2
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
```

It seems all tests are passed with java11. I think no problem at this point. What do you think?

gaul:
> I don't have ubuntu environment, so I cannot reproduce it. However, s3proxy.jar also works with java11 though it shows some warnings. IIRC, this behavior is changed since java9.

S3Proxy should work with Java versions 7-11 so please ignore the warning.  jclouds has a dependency on Guice which will take some time to upgrade...

.travis.yml
```diff
@@ -1,14 +1,10 @@
 language: go
 sudo: required
-dist: trusty
-addons:
-  apt_packages:
-   - openjdk-7-jre-headless
+dist: xenial
```
gaul: Upgrading to Ubuntu 16.04 seems reasonable since 14.04 will reach end of life next month.

test/run-tests.sh
```diff
@@ -20,6 +20,7 @@ trap cleanup EXIT
 rm -Rf /tmp/s3proxy
 mkdir -p /tmp/s3proxy
 
+java -version
```
gaul: Needed?

t2y: It is not so important. However, java in transition now. I added it since I wanted to confirm test behavior and java version on travis-ci worker when I would update travis.yaml.

kahing:
thanks @t2y

PR #416, opened 2019-04-19
list gridscale among tested providers

fkr:
I just tested our (gridscale) object storage with goofys and it works just fine.


fkr:
sure. Sorry, I somehow did not pay attention to that. ;)

kahing:
Sorry for the late response. I condensed the ceph listings since they should all work the same way. At some point it may not make sense to keep adding ceph providers but we will cross that bridge when we get there.

PR #420, opened 2019-05-23
Add assume role functionality

smartynoah:
This change allows for cross-account IAM roles via the STS AssumeRole mechanism.

This change would resolve the issue raised by a user in the following comment:
https://github.com/kahing/goofys/issues/205#issuecomment-484461906

Rather that supporting a full parse of the .aws/config file to scrape out the role to be assumed, I simply added a command line argument to specify the role ARN.

PR #428, opened 2019-06-19
Use the default credentials chain to when loading a shared AWS config

smartynoah:
This change allows the shared config to resolve a usable credentials provider by leveraging the SDK's default credentials chain.

With this change I could drop my previous PR (#420)

### DIFF
--- a/api/common/conf_s3.go
+++ b/api/common/conf_s3.go
@@ -113,6 +113,9 @@ func (c *S3Config) ToAwsConfig(flags *FlagStorage) (*aws.Config, error) {
 			if err != nil {
 				return nil, err
 			}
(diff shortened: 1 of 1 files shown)


kahing:
Unfortunately that doesn't work for people who use the api to pre-create a Credentials object. Have you tried https://github.com/kahing/goofys/pull/420#issuecomment-503415676 ? I believe the default fallback (when `Credentials` is nil) should be the default credentials chain. I tried locally to remove my `~/.aws/credentials` and seems like that's the case (although I haven't tried to test this end to end)

smartynoah:
The suggested improvement of #420 works.  Closing this.

PR #434, opened 2019-06-30
replace golang.org/x/net/context with internal context

vager:


internal/goofys_test.go
```diff
@@ -36,7 +36,7 @@ import (
 	"testing"
 	"time"
 
-	"golang.org/x/net/context"
+	"context"
```
gaul: Should this sort into the main imports above?  Repeated below.

vager: "github.com/kahing/goofys/api/common"  ?

kahing: This is probably a personal taste thing, but I usually order my imports:

1. internal imports
2. go built-in modules
3. external modules

not a big deal though

PR #437, opened 2019-07-05
Removed selecting storage backend by URL-scheme.

kacejot:
Removed selecting storage backend by URL-scheme. 
Added flag for selecting storage backend.

The reason of this changes is that URL-scheme backend detecting logic was working incorrect. 

To detect correct storage backend URL-scheme is used. So on `wasb://` I will have azure blob storage, on `s3://` I will have S3 Bucket etc.

Azure Blob Storage has next wasb request format: `wasb://mycontainer@myaccount.blob.core.windows.net`
If I pass bucket of such format, goofys parses myaccountblob.core.windows.net as bucket name, because it is specified as a host in the url. So goofys expects something like `wasb://mycontainer` to work correct, but that is not valid wasb url, because container is a part of userinfo, not the host.

On the other hand, to work with s3 we can specify no scheme, just bucket name and it will work.
The best way is to bring these methods to the same format - only bucket names without urls, just specify backend implicitly.

So I propose add --storage-backend flag to choose between s3 bucket, azure blob storage and data lake.

### DIFF
--- a/doc/backends/azure-blob-storage.md
+++ b/doc/backends/azure-blob-storage.md
@@ -1,0 +1,49 @@
+### Goofys with Azure Blob Storage
+
+Goofys is already can work with Azure Blob Storage out of the box.
+To achive successful mount of azblob container to your fs, you need to follow next simple steps:
+
+- Get account name and blob container name from your Azure Portal
+- Get the Access key for your account from Azure Portal like described [here](https://help.bittitan.com/hc/en-us/articles/115008109327-How-do-I-get-an-access-key-for-Azure-Blob-storage- "here").
+- You can put account and key values to the config file like this:
+
+```ini
+[storage]
+account = "myblobstorage"
+key = "TZR0DqSYXTVLJRoWUlHnd80qrMJTUW2nBlv1kfaz6gG13OjZpQtKt6GvpgY8JrYsKAUxbWHTHc4KU4Wps5IhYy=="
+```
+And put the file here: ~/.azure/config
+Or instead of this step you can make sure that your shell has these environment variables set before starting goofys:
+
+```sh
+export AZURE_STORAGE_ACCOUNT="yblobstorage"
+export AZURE_STORAGE_ACCESS_KEY="ZR0DqSYXTVLJRoWUlHnd80qrMJTUW2nBlv1kfaz6gG13OjZpQtKt6GvpgY8JrYsKAUxbWHTHc4KU4Wps5IhYy=="
+```
+- Install azure-cli using [this instruction](https://docs.microsoft.com/en-us/cli/azure/install-azure-cli?view=azure-cli-latest "this instruction").
+- Use azure-cli to authenticate:
+
+```bash
+# login first
+az login
+
+# list all subscribtions and select the needed one
+az account list
+
+# select current subscription (get its id from previous step)
+az account set --subscription <name or id>
+
+# write auth data to our azure folder (can take several minutes)
+az ad sp create-for-rbac --sdk-auth > ~/.azure/auth
+```
+- Add path to auth file to AZURE_AUTH_LOCATION env variable:
+
+``` bash
+export AZURE_AUTH_LOCATION="~/.azure/auth"
+```
+- Now we have all necessary info to start the goofys:
+
+```bash
+goofys --storage-backend azure-blob-storage <container-name> path/to/mount
+```
+
+Now everything should work.
--- a/internal/flags.go
+++ b/internal/flags.go
@@ -107,6 +107,12 @@ func NewApp() (app *cli.App) {
 			},
 
 			cli.StringFlag{
+				Name: "storage-backend",
+				Usage: "Select storage backend: s3, azure-blob-storage, " +
+					"azure-data-lake. (default: s3)",
+			},
+
+			cli.StringFlag{
 				Name: "cache",
 				Usage: "Directory to use for data cache. " +
 					"Requires catfs and `-o allow_other'. " +
@@ -328,6 +334,7 @@ func PopulateFlags(c *cli.Context) (ret *FlagStorage) {
 		HTTPTimeout:  c.Duration("http-timeout"),
 
 		// Common Backend Config
+		Backend:        c.String("storage-backend"),
 		Endpoint:       c.String("endpoint"),
 		UseContentType: c.Bool("use-content-type"),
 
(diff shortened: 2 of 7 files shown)


kacejot:
There is a FAIL in the test TestWriteLargeFiles, but I have run [all the tests](https://travis-ci.com/kacejot/goofys/builds/118114687) on my fork and passed them.

Could you please rerun it?

kahing:
Instead of this change, what if we make `wasb://mycontainer` and `wasb://mycontainer@myaccount.blob.core.windows.net` both work? That seems to be a much smaller change.

doc/backends/azure-blob-storage.md
```diff
@@ -0,0 +1,49 @@
+### Goofys with Azure Blob Storage
+
+Goofys is already can work with Azure Blob Storage out of the box.
+To achive successful mount of azblob container to your fs, you need to follow next simple steps:
+
+- Get account name and blob container name from your Azure Portal
+- Get the Access key for your account from Azure Portal like described [here](https://help.bittitan.com/hc/en-us/articles/115008109327-How-do-I-get-an-access-key-for-Azure-Blob-storage- "here").
+- You can put account and key values to the config file like this:
```
kahing: reference https://docs.microsoft.com/en-us/cli/azure/azure-cli-configuration?view=azure-cli-latest#cli-configuration-values-and-environment-variables . Also mention that goofys follow azure cli's configuration convention.

doc/backends/azure-blob-storage.md
```diff
@@ -0,0 +1,49 @@
...
+
+```ini
```
kahing: I'd probably change it to the same format as existing instruction for AWS:

```ini
$ cat ~/.azure/config
storage]
account = "myblobstorage"
key = "MY-STORAGE-KEY"
$ $GOPATH/bin/goofys wasb://container <mountpoint>
$ $GOPATH/bin/goofys wasb://container:prefix <mountpoint> # if you only want to mount objects under a prefix
```

internal/flags.go
```diff
@@ -328,24 +334,41 @@ func PopulateFlags(c *cli.Context) (ret *FlagStorage) {
...
 		// Common Backend Config
+		Backend:        c.String("storage-backend"),
 		Endpoint:       c.String("endpoint"),
 		UseContentType: c.Bool("use-content-type"),
 
 		// Debugging,
 		DebugFuse:  c.Bool("debug_fuse"),
 		DebugS3:    c.Bool("debug_s3"),
 		Foreground: c.Bool("f"),
+
```
kahing: do we need this?

internal/goofys.go
```diff
@@ -96,61 +96,65 @@ var log = GetLogger("main")
...
-
-func ParseBucketSpec(bucket string) (spec BucketSpec, err error) {
-	if strings.Index(bucket, "://") != -1 {
-		var u *url.URL
-		u, err = url.Parse(bucket)
+// FillBackendConfig creates config backend in the flags due to
+// the backend-storage flag
+func FillBackendConfig(bucketName string, flags *FlagStorage) error {
+	switch flags.Backend {
+	case "azure-data-lake":
```
kahing: change this to `azure-data-lake-v1`, there's also `azure-data-lake-v2` that's going to be added later

kahing:
I made the change to allow full wasb URIs and also added instruction here: https://github.com/kahing/goofys/blob/master/README-azure.md . Could you see if that works for you?

doc/backends/azure-blob-storage.md
```diff
@@ -0,0 +1,49 @@
...
+[storage]
+account = "myblobstorage"
+key = "TZR0DqSYXTVLJRoWUlHnd80qrMJTUW2nBlv1kfaz6gG13OjZpQtKt6GvpgY8JrYsKAUxbWHTHc4KU4Wps5IhYy=="
```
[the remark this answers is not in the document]
kacejot: :sweat_smile:  Yeah, it is random generated string

doc/backends/azure-blob-storage.md
```diff
@@ -0,0 +1,49 @@
+### Goofys with Azure Blob Storage
+
+Goofys is already can work with Azure Blob Storage out of the box.
+To achive successful mount of azblob container to your fs, you need to follow next simple steps:
+
+- Get account name and blob container name from your Azure Portal
+- Get the Access key for your account from Azure Portal like described [here](https://help.bittitan.com/hc/en-us/articles/115008109327-How-do-I-get-an-access-key-for-Azure-Blob-storage- "here").
```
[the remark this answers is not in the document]
kacejot: Will do

doc/backends/azure-blob-storage.md
```diff
@@ -0,0 +1,49 @@
+### Goofys with Azure Blob Storage
```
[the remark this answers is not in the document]
kacejot: I think, the wiki page is great place for such instructions.

doc/backends/azure-blob-storage.md
```diff
@@ -0,0 +1,49 @@
...
+
+Goofys is already can work with Azure Blob Storage out of the box.
```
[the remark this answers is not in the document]
kacejot: Will do

doc/backends/azure-blob-storage.md
```diff
@@ -0,0 +1,49 @@
...
+[storage]
+account = "myblobstorage"
+key = "TZR0DqSYXTVLJRoWUlHnd80qrMJTUW2nBlv1kfaz6gG13OjZpQtKt6GvpgY8JrYsKAUxbWHTHc4KU4Wps5IhYy=="
+```
+And put the file here: ~/.azure/config
+Or instead of this step you can make sure that your shell has these environment variables set before starting goofys:
+
+```sh
+export AZURE_STORAGE_ACCOUNT="yblobstorage"
+export AZURE_STORAGE_ACCESS_KEY="ZR0DqSYXTVLJRoWUlHnd80qrMJTUW2nBlv1kfaz6gG13OjZpQtKt6GvpgY8JrYsKAUxbWHTHc4KU4Wps5IhYy=="
```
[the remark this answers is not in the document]
kacejot: I will check, if it works

doc/backends/azure-blob-storage.md
kacejot: Okay

doc/backends/azure-blob-storage.md
kacejot: Does it already work? I will check anyway

internal/flags.go
kacejot: Yes, I have added all the flags user can pass as args to one structure

internal/goofys.go
kacejot: Okay

kacejot:
Hi, I was on a weekend so I wasn't able to apply edit for this PR.
I will check

kacejot:
@kahing 
Due to [this instruction](https://github.com/kahing/goofys/blob/master/README-azure.md) I think we don't really need separate flag for storage backend selection anymore. There should be only one way to select backend to be more consistent.
As far as you support several backends I suggest you to remove default s3 storage backend, because s3 bucket seems to be more valuable in this case. So all the users should use URL scheme to specify the concrete storage backend.

I have one last question: why do you use URL scheme instead of separate flag for storage backend selection? You mentioned something related with `fstab`

P.S. This PR can be closed

kahing:
removing the default for s3 is not an option since that would break existing command invocations and `fstab`.

I prefer URL scheme over separate flag for 2 reasons:

1. shorter, which is always nice but more so in `fstab` when everything has to be on one line without space
2. goofys now has internal support for mounting multiple buckets and they can be from different backends. At some point this will be exposed to command line and using url scheme makes this easier

kahing:
also, could I ask how you are using goofys on azure? I am surprised that you found it before I even added it to readme

kacejot:
@kahing
I just wanted to optimize work with azure. blobfuse was too slow

kahing:
let me know your evaluation results and any feedbacks! azure support is quite new and I plan to make a new release soon

Issue #371, opened 2018-10-16
Reproducible release build?


mengxr:
Is there a way to reproduce the official releases? I checked `dep` and it generates the following `Gopkg.toml` based on the current master. @kahing Did you keep a copy of the dependent library versions when you made the official release?

~~~
[[constraint]]
  name = "github.com/aws/aws-sdk-go"
  version = "1.15.56"

[[constraint]]
  branch = "master"
  name = "github.com/jacobsa/fuse"

[[constraint]]
  branch = "master"
  name = "github.com/jinzhu/copier"

[[constraint]]
  name = "github.com/kahing/go-xattr"
  version = "1.1.1"

[[constraint]]
  branch = "master"
  name = "github.com/kardianos/osext"

[[constraint]]
  name = "github.com/sevlyar/go-daemon"
  version = "0.1.4"

[[constraint]]
  name = "github.com/shirou/gopsutil"
  version = "2.18.7"

[[constraint]]
  name = "github.com/sirupsen/logrus"
  version = "1.1.1"

[[constraint]]
  name = "github.com/urfave/cli"
  version = "1.20.0"

[[constraint]]
  branch = "master"
  name = "golang.org/x/net"

[[constraint]]
  branch = "v1"
  name = "gopkg.in/check.v1"

[prune]
  go-tests = true
  unused-packages = true
~~~

kahing:
agreed that this is desired but go keeps switching the de facto tool on dependencies. Lets wait for https://github.com/golang/go/wiki/Modules to be finalized

PR #462, opened 2019-09-13
Switch to using go modules for building.

georgyo:
Using git submodules requires having environment variables being setup locally. The following patch set allows you to just run ```go build``` and have go do the right thing.

There is also fix as go.uuid.NewV4 does not return an error any more.


kahing:
This is great! Looks like CI currently fails though: https://travis-ci.org/kahing/goofys/jobs/584784568

ref https://github.com/kahing/goofys/issues/371

georgyo:
So, I don't understand what is going on with the Azure Datalake tests.  I even downgraded all the azure modules to the exact version you had before.

The error is as follow, but the test does seem like it should be getting that cleanly from the secure environment variables. Though I don't know enough about both travis-ci and azure to debug this deeper.
```
START: goofys_test.go:2791: GoofysTest.TestAzureDirBlob
START: goofys_test.go:317: GoofysTest.SetUpTest
goofys_test.go:443:
    t.Assert(err, IsNil)
... value *errors.errorString = &errors.errorString{s:"failed to get SPT from client credentials: parameter 'clientID' cannot be empty"} ("failed to get SPT from client credentials: parameter 'clientID' cannot be empty")
```

georgyo:
It seems those tests are actually broken, looking at other merge requests which don't even go close to those azure code paths. Cleaned up the commit history so this can be merged.

kahing:
Oh right, the secure variables are only available to master branch (otherwise anyone can create a PR to dump out my credentials). I will look into testing this soonish

georgyo:
@kahing Just seeing if this is still something you would be interested in merging.

NefixEstrada:
What's the status of this? The whole project is impossible to use if this doesn't get merged

svaroqui:
Hi any progress on that merge ? I'm intrested in it because not building when including
import (
	goofys "github.com/kahing/goofys/api"
	common "github.com/kahing/goofys/api/common"

	"fmt"
	"context"
)

svaroqui:
What's the status of this? The whole project is impossible to use if this doesn't get merged

georgyo:
I would need to refresh this for it to be usable again, but I don't think @kahing has much interest in merging this type of change.

svaroqui:
Ok did this 
replace github.com/kahing/goofys => github.com/georgyo/goofys v0.21.0

but it did not make it 
# github.com/kahing/goofys/internal
../../../../pkg/mod/github.com/georgyo/goofys@v0.21.0/internal/backend_adlv1.go:112:9: assignment mismatch: 2 variables but "github.com/satori/go.uuid".NewV4 returns 1 values
../../../../pkg/mod/github.com/georgyo/goofys@v0.21.0/internal/backend_adlv1.go:527:15: assignment mismatch: 2 variables but "github.com/satori/go.uuid".NewV4 returns 1 values

kahing:
e3748e3cdfe2d2f5ca775e6370020ff863ad6599

PR #469, opened 2019-10-12
use go mod

fengpf:
upgrade go version to 1.13


georgyo:
This does not build...

Alternative implementation here: #462

The major advantage to not rewriting all the import lines is that the version is #462 is still `go get` able

kahing:
e3748e3cdfe2d2f5ca775e6370020ff863ad6599

PR #471, opened 2019-10-18
"only sequential writes supported" error in case of zero data/file hole

jmpq:
'cp' might skip over the zero data/file hole when copying a file to goofys,
causing "only sequential writes supported" error because offset != fh.nextWriteOffset.

Fix the issue by filling the range [offset, fh.nextWriteOffset) with zero.


kahing:
there's WIP work here as well: https://github.com/kahing/goofys/commits/sparse

also depends on https://github.com/jacobsa/fuse/pull/66 which cp may need depends on where the sparse region is

PR #485, opened 2020-01-14
Get *resp.ContentLength error, when dir is deep.

yvgan818:
When object is like /a/b/c/d... .Use command “cd MOUNT_DIR/a/b/c” while return an empty Content-Length. Get *resp.ContentLength will be error.
ceph version is 12.2.10.


kahing:
Why would `Content-Length` be empty?

kahing:
this sounds like a ceph bug, have you file an issue there?

PR #502, opened 2020-03-04
wait for mpuid until got in waitForCreateMPU

kungf:
if mpuId is nil in mpuPart, will deadlock, because fh.mpuWG.Add(1)
called before mpuPart, it can not wait itself until defer mpuWG.Done().
we can wait mpuid in waitForCreateMPU until got it.

### DIFF
--- a/internal/file.go
+++ b/internal/file.go
@@ -151,15 +151,6 @@ func (fh *FileHandle) mpuPart(buf *MBuf, part uint32, total int64) {
 		fh.mpuWG.Done()
 	}()
 
-	// maybe wait for CreateMultipartUpload
-	if fh.mpuId == nil {
-		fh.mpuWG.Wait()
-		// initMPU might have errored
-		if fh.mpuId == nil {
-			return
-		}
-	}
-
 	err := fh.mpuPartNoSpawn(buf, part, total, false)
 	if err != nil {
 		if fh.lastWriteError == nil {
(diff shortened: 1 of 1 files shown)


kahing:
sorry I don't understand the issue here or the change you are trying to make. does it address a real issue or a perceived code quality issue?

kungf:
In the code below, 
1. fh.mpuWG.Add(1)
2. then go  fh.mpuPart()
3. if fh.mpuId  is nil, then fh.mpuWG.Wait(), here will deadlock if it first mpuWG add(1) then wait() mpuWG to be zero, before Done() release

```
func (fh *FileHandle) uploadCurrentBuf(parallel bool) (err error) {
	err = fh.waitForCreateMPU()
        ......
	if parallel {
		fh.mpuWG.Add(1)
		go fh.mpuPart(buf, part, fh.nextWriteOffset)
                 /--- this is part of mpuPart func
	              if fh.mpuId == nil {
		              fh.mpuWG.Wait()
                ---/
	} else {
        ......
	}
	return
}
```
there is a simple test case:
`
package main
import (
	"fmt"
	"sync"
)

var wg sync.WaitGroup
func test() {
	wg.Wait()
	fmt.Println("test")
	wg.Done()
}

func main() {
	wg.Add(1)
	go test()
	wg.Wait()
}
`

PR #516, opened 2020-05-01
Add support for mounting S3 Access Points

callumsmits:
This PR adds support for mounting S3 Access Points. The following examples now mount the Access point:

```bash
# Mount without a prefix
$ ./goofys arn:aws:s3:us-west-2:123456789012:accesspoint/test-access-point /tmp/goofys-mount

# Mount with a prefix
$ ./goofys arn:aws:s3:us-west-2:123456789012:accesspoint/test-access-point:sample-prefix /tmp/goofys-mount
```

### DIFF
--- a/internal/goofys.go
+++ b/internal/goofys.go
@@ -34,8 +34,9 @@ import (
 	"github.com/jacobsa/fuse/fuseops"
 	"github.com/jacobsa/fuse/fuseutil"
 
-	"github.com/sirupsen/logrus"
 	"net/http"
+
+	"github.com/sirupsen/logrus"
 )
 
 // goofys is a Filey System written in Go. All the backend data is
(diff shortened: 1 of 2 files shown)


elgalu:
Hi @callumsmits could you solve the conflicts? Thanks!

internal/goofys.go
```diff
@@ -34,8 +34,9 @@ import (
 	"github.com/jacobsa/fuse/fuseops"
 	"github.com/jacobsa/fuse/fuseutil"
 
-	"github.com/sirupsen/logrus"
```
elgalu, line 37: why changing the order here?

callumsmits:
Hi @elgalu, I can if this is still wanted - S3 recently launched S3 Access Point Aliases (https://aws.amazon.com/about-aws/whats-new/2021/07/amazon-s3-access-points-aliases-allow-application-requires-s3-bucket-name-easily-use-access-point/). These allow access to an Access Point as though it was a bucket. I tested it out with the 0.24.0 goofys release and it works as expected:
```bash
$ goofys test-access-point-bmc9riew8cdh3hzxd61frh9bsubhsusw2a-s3alias /tmp/goofys-mount && ls /tmp/goofys-mount
sample-prefix test.txt
```
Given that goofys can now mount Access Points, I'm not sure its worth adding extra complexity to allow access via the Access Point ARN?

elgalu:
Good point, this is not needed then. Thanks!

PR #521, opened 2020-05-21
update aws-sdk-go to v1.31.2

bradyforcier1:


api/common/conf_s3.go
```diff
@@ -73,6 +73,7 @@ func (c *S3Config) ToAwsConfig(flags *FlagStorage) (*aws.Config, error) {
 	awsConfig := (&aws.Config{
 		Region: &c.Region,
 		Logger: GetLogger("s3"),
+		Retryer: client.DefaultRetryer{NumMaxRetries: 20,},
```
bradyforcier1, line 76: Open to making this a flag, but it might just allow people to get in trouble by playing with it

PR #524, opened 2020-06-10
Add GCS Backend to Goofys - ReadOnly Access

deka108:


internal/backend_gcs.go
```diff
@@ -0,0 +1,276 @@
...
+		ContentType: &attrs.ContentType,
+		IsDirBlob: strings.HasSuffix(param.Key, "/"),
+		Metadata: metadata,
+	}, nil
+}
+
+func mapGCSMetadataToBackendMetadata(m map[string]string) map[string]*string {
+	newMap := make(map[string]*string)
+
+	for k, v := range m {
```
dotslash, line 126: why toLower?

internal/backend_gcs.go
```diff
@@ -0,0 +1,276 @@
...
+	}
+
+	return newMap
+}
+
+func (g *GCSBackend) ListBlobs(param *ListBlobsInput) (*ListBlobsOutput, error) {
+	// TODO: Ignore object versions from listing
+	query := storage.Query{}
+
+	if param.Prefix != nil {
```
dotslash, line 138: nilStr handles the nil check

internal/backend_gcs.go
```diff
@@ -0,0 +1,276 @@
...
+	}
+	if param.Delimiter != nil {
+		query.Delimiter = nilStr(param.Delimiter)
+	}
+
+	prefixes := make([]BlobPrefixOutput, 0)
+	items := make([]BlobItemOutput, 0)
+
+	it := g.bucket.Objects(context.Background(), &query)
+	g.logger.Debugf("LIST Prefix = %s Delim = %s", nilStr(param.Prefix), nilStr(param.Delimiter))
```
dotslash, line 149: user query.Prefix and query.Delimiter

internal/backend_gcs.go
```diff
@@ -0,0 +1,276 @@
...
+	for {
+		attrs, err := it.Next()
+		if err == iterator.Done {
+			break
```
dotslash, line 153: break early based on `param.MaxKeys`

internal/backend_gcs.go
```diff
@@ -0,0 +1,276 @@
...
+				Size: uint64(attrs.Size),
+				StorageClass: &attrs.StorageClass,
+			})
+		}
+	}
+
+	return &ListBlobsOutput{
+		Prefixes: prefixes,
+		Items: items,
+		NextContinuationToken: nil,
```
dotslash, line 175: Add a note why this is never truncated.

internal/backend_gcs.go
```diff
@@ -0,0 +1,276 @@
...
+		return nil, err
+	}
+	g.logger.Debugf("GET ATTRS %v = %s ", param.Key, getResponseStatus(err))
+
+	// Get Object reader
+	var rc *storage.Reader
+	if param.Start != 0 || param.Count != 0 {
+		rc, err = obj.NewRangeReader(ctx, int64(param.Start), int64(param.Count))
+	} else {
+		rc, err = obj.NewReader(ctx)
```
dotslash, line 214: What if the blob of the size 100GB?


We might always want the range reader

internal/backend_gcs.go
```diff
@@ -0,0 +1,276 @@
...
+		HeadBlobOutput: HeadBlobOutput{
+			BlobItemOutput: BlobItemOutput{
+				Key: &attrs.Name,
+				ETag: &attrs.Etag,
+				LastModified: &rc.Attrs.LastModified,
+				Size: uint64(rc.Attrs.Size),
+				StorageClass: &attrs.StorageClass,
+			},
+			ContentType: &rc.Attrs.ContentType,
+			//Metadata: metadata,
```
dotslash, line 235: why no metadata?

internal/backend_gcs_test.go
```diff
@@ -0,0 +1,114 @@
...
+	viper.SetConfigFile(env.GetEnv("CONFIG_FILE", true))
+
+	err := viper.ReadInConfig() // Find and read the config file
+	if err != nil { // Handle errors reading the config file
+		panic(fmt.Errorf("Fatal error config file: %s \n", err))
+	}
+}
+
+
+func TestMain(m *testing.M) {
```
dotslash, line 26: Lets use one testing framework/lib for the entire project. (check.v1)

internal/backend_gcs_test.go
```diff
@@ -0,0 +1,114 @@
...
+	gcsBackend, _ = getGcsBackend()
+	exitVal := m.Run()
+	log.Println("Do stuff AFTER the tests!")
+
+	os.Exit(exitVal)
+}
+
+func TestGCSConfig_WithoutCredentials(t *testing.T){
+	val := os.Getenv("GOOGLE_APPLICATION_CREDENTIALS")
+	os.Unsetenv("GOOGLE_APPLICATION_CREDENTIALS")
```
dotslash, line 38: Do these tests always run serially?

internal/backend_gcs_test.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+func TestGCSConfig_WithCredentials(t *testing.T){
+	_, err := common.NewGCSConfig()
+	assert.Nil(t, err)
+}
+
+func getGcsConfig(bucket string) (*common.GCSConfig, error) {
+	spec, err := ParseBucketSpec(bucket)
+	if err != nil {
```
dotslash, line 53: order: helper functions first; tests later

internal/backend_gcs_test.go
```diff
@@ -0,0 +1,114 @@
...
+	config.Bucket = spec.Bucket
+	if config.Prefix != "" {
+		config.Prefix = spec.Prefix
+	}
+	return config, nil
+}
+
+
+func TestGCSBackend_CreateBackend(t *testing.T){
+	config, _ := getGcsConfig(viper.GetString("goofys.gcs.bucketWithPermission"))
```
dotslash, line 70: where is this config (for viper)

internal/backend_gcs_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+	if err != nil {
+		fmt.Println(err)
+	}
+	fmt.Println(gcsBackend)
+}
+
+func TestGCSBackend_HeadBlob(t *testing.T) {
+	testCases := []struct {
+		input   string
```
dotslash, line 81: define a struct. this is hard to read

internal/flags.go
```diff
@@ -257,6 +257,11 @@ func NewApp() (app *cli.App) {
 				Usage: "Enable S3-related debugging output.",
 			},
 
+			cli.BoolFlag{
+				Name:  "debug_cloud",
+				Usage: "Enable Cloud Storage-related debugging output.",
```
dotslash, line 262: may be mark debug_s3 as deprecated ?

internal/string_test.go
```diff
@@ -0,0 +1,51 @@
+package internal
+
+import (
+	"fmt"
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func TestParseBucketSpec(t *testing.T){
```
dotslash, line 9: rename the file to something better. bucket_spec_parse_test?

internal/string_test.go
```diff
@@ -0,0 +1,51 @@
...
+	assert := assert.New(t)
+
+	testCases := []struct{
+		input string
+		expected BucketSpec
+	}{
```
dotslash, line 15: This is a nice test. Add more cases or leave a todo(kahing) 😛

internal/string_test.go
```diff
@@ -0,0 +1,51 @@
...
+
+	for _, tc := range testCases {
+		spec, err := ParseBucketSpec(tc.input)
+		if err != nil {
+			log.Fatal(err)
+		}
+		fmt.Println(fmt.Sprintf("Input: %s, Scheme: %s, Bucket: %s, Prefix: %s",
+			tc.input, spec.Scheme, spec.Bucket, spec.Prefix))
+		assert.Equal(tc.expected.Scheme, spec.Scheme)
+		assert.Equal(tc.expected.Bucket, spec.Bucket)
```
dotslash, line 47: Is there a way to deep equals 2 structs?

internal/gcs_test.go
```diff
@@ -0,0 +1,252 @@
...
+func getGcsBackend() (*GCSBackend, error) {
+	config, _ := getGcsConfig(viper.GetString("goofys.gcs.bucketWithPermission"))
+	gcsBackend, err := NewGCS(config)
+
+	return gcsBackend, err
+}
+
+var gcsBackend *GCSBackend;
+
+func TestGCS_BlobExist(t *testing.T) {
```
dotslash, line 25: what are you testing in the file? Whether you are using GCS apis correctly or not?

These tests rely on existnace of a gcs bucket with a particular set of files. To checkin this we need to find a way so that anyone without credentials can run the test.

api/common/conf_gcs.go
```diff
@@ -0,0 +1,28 @@
...
+	Bucket string
+	Prefix string
+	//Scope
+}
+
+func NewGCSConfig() (*GCSConfig, error){
+	ctx := context.Background()
+
+	// v0.1: only allows authenticated user and read only acccess
+	//credentials, err := google.FindDefaultCredentials(ctx)
```
dotslash, line 19: gofmt

internal/backend_gcs.go
```diff
@@ -0,0 +1,276 @@
+package internal
+
+import (
+	"cloud.google.com/go/storage"
```
dotslash, line 4: I forgot. Did we decide to only use the "polished" google api?

Can the polished and the raw api be used side-by-side?

deka108: Polished API (aka client API). I think we can use the raw and polished api at the same time because the polished API is implemented on top of the raw API. I'd recommend to limit the usage of raw API because they can change at anytime and it can be a breaking change (not backward compatible)

internal/backend_gcs.go
deka108: i think i just followed https://github.com/kahing/goofys/blob/45b8d78375af1b24604439d2e60c567654bcdf88/internal/backend_s3.go#L331 where they tried to normalize the metadata keys into having lowercase keys

PR #535, opened 2020-07-10
Make ReadAhead related const into optional arguments.

maobaolong:


maobaolong:
@kahing Please take a look at this PR. Thanks.

kahing:
could you explain why you want these?

maobaolong:
@kahing Thanks for your reply.

This PR aimed to supply a way to custom the readAhead chunkSize and readAhead maxSize. For some higher performance request case, we want to read ahead more data, so we want customized the readAhead chunkSize and readAhead maxSize.

maobaolong:
Hi @kahing , how about your idea? We use this feature in our product env for months, it can helps us to customize the readAhead chunkSize and readAhead maxSize.

PR #537, opened 2020-07-17
Add a partSize parameter.

maobaolong:
By this parameter, we can specify a customized partSize.


internal/file.go
```diff
@@ -187,11 +187,11 @@ func (fh *FileHandle) partSize() uint64 {
 	var size uint64
 
 	if fh.lastPartId < 1000 {
-		size = 5 * 1024 * 1024
+		size = fh.inode.fs.flags.MPUPartSize
```
gaul, line 190: This behavior may confuse users.  Instead, how about using a fixed size when provided by the flag and this gradually increasing size when not provided?

kahing:
Could you explain what's the motivation of configuring a part size?

maobaolong:
@kahing Now Ozone use Goofys as a fuse interface, Goofys talk to Ozone s3gateway now, but there are a performance issue when the partSize is not suitable to Ozone chunkSize, now Ozone chunk is 4MB, and goofys partSize is 5MB, so it will cost two time read Chunk when read file, If we set the partSize to 8MB, when read this part, two time read chunk can read 8MB from ozone datanode, with this feature, we can improve 3x read performance.

internal/file.go
maobaolong: OK, we can further discussing this after we accept this proposal

PR #543, opened 2020-08-13
Add pid file support

zachomedia:
This PR adds support for writing a PID file. This lets us clean up our Kubernetes flexvol mounts (https://github.com/StatCan/s3-fuse-flex-volume) since we can track the pid of the goofys instance we launch.

### DIFF
--- a/api/common/config.go
+++ b/api/common/config.go
@@ -26,6 +26,8 @@ import (
 )
 
 type FlagStorage struct {
+	PidFile string
+
 	// File system
 	MountOptions      map[string]string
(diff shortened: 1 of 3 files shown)


kahing:
Thanks @zachomedia . Could you put this next to `-f` under `misc options`. Also explain that this doesn't work with `-f`. Maybe something like: 

> Write the background process id to this file. Does nothing if `-f` is used`.

zachomedia:
@kahing Thanks, I've made the requested change.

PR #547, opened 2020-08-21
Memory optimization when using aws s3

ebressler:
The changes here make it so that there are no longer references being
held to the s3 objects allocated during XMLToStruct in the s3 libraries.

In my testing, when doing a listObjectsV2 against ~383,000 objects, this
change saves about 1/2 of the memory uses. (1gb of use without these
changes and ~500mb with them)

### DIFF
--- a/internal/flags.go
+++ b/internal/flags.go
@@ -86,7 +86,7 @@ func NewApp() (app *cli.App) {
 
 	app = &cli.App{
 		Name:     "goofys",
-		Version:  "0.23.1-" + VersionHash,
+		Version:  "0.24.0-" + VersionHash,
 		Usage:    "Mount an S3 bucket locally",
 		HideHelp: true,
 		Writer:   os.Stderr,


kahing:
Thanks @ebressler . Could you explain why this matters? Esp since because for the most part we don't keep the resulting `BlobItemOutput` around? Is it because of https://github.com/kahing/goofys/blob/master/internal/dir.go#L514 ?

The reason I am asking is it's likely that other backend would have the same problem too, and it'd be nice to fix it once instead of each `backend_*` fixing it each time

ebressler:
Hey @kahing ,thanks for the question. I believe the problem is that since memory pointers were used to the original strings, it kept a reference all the way back to the S3 object. Since the inodes get put into the cache and never get purged from the cache unless they are accessed, they reference count never goes to 0. I am sure that the same problem would be there for other backends, but I don't have a way to check that right now easily.  I am pretty sure it is not just that basename slice you point to. However, it could be that plus inside of SetFromBlobItem (https://github.com/kahing/goofys/blob/master/internal/dir.go#L532) that this is happening. I will try making the change there instead and see if that solves it because I didn't like that I had to sprinkle 3 different ways of doing the same thing across the file because the data filling BlobItemOuput was gathered slightly differently in each case.

Anyway, for more detail before you had a chain that looked something like:

inode->BlobItemOuput->ListObjectV2Output->s3.ListObjectsOutput where there was always at least 1 pointer reference to a part of a string in a struct going all the way back.

This keeps the s3ListObjectsOutput around. Showed up very clearly in a pprof.

Before:
```
File: __debug_bin
Build ID: 08715b78007cf3be19942153eb7a6cb9d8ff35ad
Type: inuse_space
Time: Aug 17, 2020 at 2:14pm (EDT)
Entering interactive mode (type "help" for commands, "o" for options)
(pprof) top
Showing nodes accounting for 738.16MB, 98.75% of 747.51MB total
Dropped 37 nodes (cum <= 3.74MB)
Showing top 10 nodes out of 27
      flat  flat%   sum%        cum   cum%
  414.56MB 55.46% 55.46%   560.56MB 74.99%  github.com/aws/aws-sdk-go/private/protocol/xml/xmlutil.XMLToStruct
  121.51MB 16.25% 71.71%   121.51MB 16.25%  github.com/aws/aws-sdk-go/private/protocol/xml/xmlutil.(*XMLNode).findNamespaces
  102.03MB 13.65% 85.36%   102.03MB 13.65%  github.com/kahing/goofys/internal.(*Inode).SetFromBlobItem
   54.51MB  7.29% 92.65%    54.51MB  7.29%  github.com/kahing/goofys/internal.NewInode
   18.50MB  2.47% 95.13%    18.50MB  2.47%  encoding/xml.(*Decoder).name
   16.56MB  2.22% 97.34%    19.29MB  2.58%  github.com/kahing/goofys/internal.(*Goofys).insertInode
    5.50MB  0.74% 98.08%     5.50MB  0.74%  encoding/xml.makeCopy
    4.50MB   0.6% 98.68%   179.32MB 23.99%  github.com/kahing/goofys/internal.(*DirHandle).ReadDir
    0.50MB 0.067% 98.75%       19MB  2.54%  encoding/xml.(*Decoder).rawToken
         0     0% 98.75%       19MB  2.54%  encoding/xml.(*Decoder).Token

```

After:
```
File: __debug_bin
Build ID: 8ee5b83b63225a26da993bb9b49755fed82fb3a2
Type: inuse_space
Time: Aug 19, 2020 at 5:19pm (EDT)
Entering interactive mode (type "help" for commands, "o" for options)
(pprof) top
Showing nodes accounting for 348.99MB, 95.02% of 367.28MB total
Dropped 56 nodes (cum <= 1.84MB)
Showing top 10 nodes out of 51
      flat  flat%   sum%        cum   cum%
  124.04MB 33.77% 33.77%   124.04MB 33.77%  github.com/kahing/goofys/internal.(*Inode).SetFromBlobItem
   67.01MB 18.24% 52.02%    67.01MB 18.24%  github.com/kahing/goofys/internal.NewInode
   43.50MB 11.84% 63.86%    49.54MB 13.49%  github.com/aws/aws-sdk-go/private/protocol/xml/xmlutil.XMLToStruct
   37.28MB 10.15% 74.01%    50.88MB 13.85%  github.com/dgraph-io/ristretto.(*lockedMap).Set
      32MB  8.71% 82.73%       32MB  8.71%  github.com/dgraph-io/ristretto.newCmRow
      16MB  4.36% 87.08%       16MB  4.36%  github.com/dgraph-io/ristretto/z.(*Bloom).Size
   13.60MB  3.70% 90.78%    13.60MB  3.70%  github.com/dgraph-io/ristretto.(*expirationMap).add
    9.56MB  2.60% 93.39%     9.56MB  2.60%  github.com/dgraph-io/ristretto.(*sampledLFU).add
       3MB  0.82% 94.20%        3MB  0.82%  github.com/aws/aws-sdk-go/private/protocol/xml/xmlutil.(*XMLNode).findNamespaces
       3MB  0.82% 95.02%   198.76MB 54.12%  github.com/kahing/goofys/internal.(*DirHandle).ReadDir
```

PR #555, opened 2020-09-11
Change to use Go Mod

ebressler:
This update will make it so that it can be used with `go mod` for vendoring so that it is no longer necessary to use submodules or even have a vendor directory. This makes it so that now to build it, you simply need to do a `go build` and it will grab everything it needs. Also, notice that it basically has an alias to ` github.com/kahing/fusego`. If that code ever merges back, then removing that replace will make it so that it is back at the mainline fusego, but otherwise people can use the fork and not run into wierd problems with masquerading ` github.com/kahing/fusego` as `github.com/jacobsa/fuse`.


ebressler:
@kahing anything you need from me on this?

ebressler:
@kahing, just checking in on this.

amine250:
Hello @kahing 

When will this be bundled into a release? 

Thanks.

PR #558, opened 2020-09-16
Add GetFullName() back

bensussman:
This is effectively the same PR as https://github.com/kahing/goofys/pull/330 which was merged back in 2018 and allowed us at Spell to wrap `goofys` into a `goofyscache` library that solves our specific use cases that `catfs` does not.

We are currently live on an old version of goofys, but would like to take advantage of the recently added Azure support and in order to upgrade we need the `GetFullName()` method to be added back.

Thanks!

### DIFF
--- a/internal/goofys.go
+++ b/internal/goofys.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"math/rand"
+	"net/http"
 	"net/url"
 	"runtime/debug"
 	"strings"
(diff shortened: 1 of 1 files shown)


bensussman:
@kahing i'm not able to add you as a "Reviewer" or "Assignee". LMK how you'd like to proceed

bensussman:
It looks like it was removed here: https://github.com/kahing/goofys/commit/e6364958dc52b9eff01009f61a5c73be287a7510

although i'm not sure what the reasoning was?

gaul:
@bensussman goofyscache sounds interesting -- can you publish this somewhere?

bensussman:
@gaul I think that can be arranged! It currently is a submodule in spell's proprietary "API" but as part of this project I have been meaning to make it its own go module, and pull it into it's own git repo. I'll work on it in the coming months.

PR #562, opened 2020-09-23
use SetLevel() instead of Level = ? to prevent data race

bensussman:
This fixes a data race!

### DIFF
--- a/api/common/logger.go
+++ b/api/common/logger.go
@@ -119,14 +119,14 @@ func GetLogger(name string) *LogHandle {
 
 	if logger, ok := loggers[name]; ok {
 		if name != "main" && name != "fuse" {
-			logger.Level = cloudLogLevel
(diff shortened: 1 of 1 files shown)


api/common/logger.go
```diff
@@ -119,14 +119,14 @@ func GetLogger(name string) *LogHandle {
 
 	if logger, ok := loggers[name]; ok {
 		if name != "main" && name != "fuse" {
-			logger.Level = cloudLogLevel
+			logger.SetLevel(cloudLogLevel)
```
gaul, line 122: `SetLevel` calls `atomic.StoreUint32` which avoids the race:

https://github.com/sirupsen/logrus/blob/42baed85eb233bfc674562af0ef6e99c45f5f384/logger.go#L357

gaul (approved):

Issue #571, opened 2020-10-26
Duplicate inode created on MkDir when executed simultaneously with same name


skuppa:
The goofy restarts on when same directory created simultaneously.  The stack trace from goofy is

```
panic: double insert of some_name/folders/37079

goroutine 11332 [running]: github.com/kahing/goofys/internal.(*Inode).insertChildUnlocked(0xc00026e900, 0xc000685710) /root/go/src/github.com/kahing/goofys/internal/handles.go:246 +0x4bd
github.com/kahing/goofys/internal.(*Goofys).insertInode(0xc000290c30, 0xc00026e900, 0xc000685710) /root/go/src/github.com/kahing/goofys/internal/goofys.go:687 +0x54
github.com/kahing/goofys/internal.(*Goofys).MkDir(0xc000290c30, 0xd3a200, 0xc000dbf050, 0xc000d422a0, 0x0, 0x0) /root/go/src/github.com/kahing/goofys/internal/goofys.go:1029 +0x1a7
github.com/jacobsa/fuse/fuseutil.(*fileSystemServer).handleOp(0xc000398ba0, 0xc0002905b0, 0xd3a200, 0xc000dbf050, 0xa2e7c0, 0xc000d422a0) /root/go/pkg/mod/github.com/jacobsa/fuse@v0.0.0-20180417054321-cd3959611bcb/fuseutil/file_system.go:154 +0xb0b
created by github.com/jacobsa/fuse/fuseutil.(*fileSystemServer).ServeOps /root/go/pkg/mod/github.com/jacobsa/fuse@v0.0.0-20180417054321-cd3959611bcb/fuseutil/file_system.go:121 +0x173
```

skuppa-veeva:
Here is PR to fix this issue: 

https://github.com/kahing/goofys/pull/572

skuppa-veeva:
@kahing I created a PR to fix this issue, could you review and let me know your feedback.

Issue #377, opened 2018-11-09
Constant goofys crash


kangaroo:
I'm getting a consistent crash on every moderate sized upload:

```
goroutine 28 [running]: 
github.com/kahing/goofys/internal.(*Inode).DeRef(0xc437888ea0, 0x1, 0xc437888ea0) 
        /home/plasma/work/src/github.com/kahing/goofys/internal/handles.go:290 +0x26d 
github.com/kahing/goofys/internal.(*Goofys).ForgetInode(0xc420165ee0, 0xa07340, 0xc45719ff20, 0xc43786d5b0, 0x89c240, 0xc43786d5b0) 
        /home/plasma/work/src/github.com/kahing/goofys/internal/goofys.go:697 +0x79 
github.com/jacobsa/fuse/fuseutil.(*fileSystemServer).handleOp(0xc4203b21e0, 0xc4203a81a0, 0xa07340, 0xc45719ff20, 0x89c240, 0xc43786d5b0) 
        /home/plasma/work/src/github.com/jacobsa/fuse/fuseutil/file_system.go:151 +0x8c4 
github.com/jacobsa/fuse/fuseutil.(*fileSystemServer).ServeOps(0xc4203b21e0, 0xc4203a81a0) 
        /home/plasma/work/src/github.com/jacobsa/fuse/fuseutil/file_system.go:119 +0x11c 
github.com/jacobsa/fuse.Mount.func1(0xa02260, 0xc4203b21e0, 0xc4203a81a0, 0xc420176018) 
        /home/plasma/work/src/github.com/jacobsa/fuse/mount.go:89 +0x3f 
created by github.com/jacobsa/fuse.Mount 
        /home/plasma/work/src/github.com/jacobsa/fuse/mount.go:88 +0x497 
```

gaul:
Which operation causes this?  Can you reproduce a test case?

silenceli:
I meet the same issue:
```
panic: Unknown inode: 2209

goroutine 40 [running]:
github.com/kahing/goofys/internal.(*Goofys).getInodeOrDie(0xc420220680, 0x8a1, 0xc422d3dec0)
	/opt/ml/bin/work/src/github.com/kahing/goofys/internal/goofys.go:421 +0x126
github.com/kahing/goofys/internal.(*Goofys).ForgetInode(0xc420220680, 0xa00a00, 0xc4230111a0, 0xc422d55000, 0x896dc0, 0xc422d55000)
[...]

Issue #353, opened 2018-08-29
Double writing causes goofys crashes


remrain:
Goofys crashes when writing the same file simultaneous.

Ex:
date >mount-test/data & date >mount-test/data & date >mount-test/data

```
2018/08/29 19:51:31.111694 s3.DEBUG HEAD mount-test/data/ = no such file or directory
2018/08/29 19:51:31.111718 fuse.DEBUG <-- LookUpInode 7 data no such file or directory
2018/08/29 19:51:31.111816 fuse.DEBUG Op 0x00000003        connection.go:481] -> Error: "no such file or directory"
2018/08/29 19:51:31.111978 fuse.DEBUG Op 0x00000002        connection.go:396] <- CreateFile (parent 7, name "data")
2018/08/29 19:51:31.112092 fuse.DEBUG Create 7 mount-test [data]
2018/08/29 19:51:31.112144 fuse.DEBUG <-- CreateFile 13 mount-test/data []
2018/08/29 19:51:31.112198 fuse.DEBUG Op 0x00000002        connection.go:479] -> OK (inode 13)
2018/08/29 19:51:31.112380 fuse.DEBUG Op 0x00000003        connection.go:396] <- CreateFile (parent 7, name "data")
2018/08/29 19:51:31.112500 fuse.DEBUG Create 7 mount-test [data]
panic: double insert of mount-test/data

goroutine 177 [running]:
github.com/kahing/goofys/internal.(*Inode).insertChildUnlocked(0xc420114240, 0xc4201d4480)
	/private/tmp/goofys-20180109-35985-h1zbtw/gopath/src/github.com/kahing/goofys/internal/handles.go:246 +0x616
github.com/kahing/goofys/internal.(*Goofys).insertInode(0xc420161ee0, 0xc420114240, 0xc4201d4480)
	/private/tmp/goofys-20180109-35985-h1zbtw/gopath/src/github.com/kahing/goofys/internal/goofys.go:687 +0x53
github.com/kahing/goofys/internal.(*Goofys).CreateFile(0xc420161ee0, 0x4891140, 0xc420262f30, 0xc4208e6000, 0x0, 0x0)
	/private/tmp/goofys-20180109-35985-h1zbtw/gopath/src/github.com/kahing/goofys/internal/goofys.go:973 +0x138
[...]

PR #572, opened 2020-10-26
Fixing the creation of multiple inode creation for the same name when…

skuppa-veeva:
Fixing the creation of multiple inode creation for the same name when MkDir executed simultaneously.


internal/goofys.go
```diff
@@ -1046,8 +1046,8 @@ func (fs *Goofys) MkDir(
 	op *fuseops.MkDirOp) (err error) {
 
 	fs.mu.RLock()
+	defer fs.mu.RUnlock()
```
skuppa-veeva, line 1049: Need to unlock after MkDir otherwise, the MkDir creates duplicate inode for same name when MkDir is called with same name.

internal/goofys.go
```diff
@@ -1056,11 +1056,7 @@ func (fs *Goofys) MkDir(
 	}
 
 	parent.mu.Lock()
-
-	fs.mu.Lock()
```
skuppa-veeva, line 1060: We don't need this lock since this lock is already acquired in the beginning of the method.

skuppa-veeva:
It is better to get `parent.mu.Lock()` before calling function `MkDir`.  A same problem exists with `CreateFile` as well.  More changes are coming with this PR.

skuppa:
> It is better to get `parent.mu.Lock()` before calling function `MkDir`. A same problem exists with `CreateFile` as well. More changes are coming with this PR.

Done!

internal/dir.go
```diff
@@ -905,9 +905,6 @@ func (parent *Inode) Create(
 
 	fs := parent.fs
 
-	parent.mu.Lock()
-	defer parent.mu.Unlock()
-
```
skuppa-veeva, line 910: We don't need to parent since it is already locked from caller.

internal/dir.go
```diff
@@ -948,9 +945,6 @@ func (parent *Inode) MkDir(
 		return
 	}
 
-	parent.mu.Lock()
-	defer parent.mu.Unlock()
-
```
skuppa-veeva, line 953: We need lock here as well, since caller already has a lock.

internal/goofys.go
```diff
@@ -1013,10 +1013,10 @@ func (fs *Goofys) CreateFile(
 	parent := fs.getInodeOrDie(op.Parent)
 	fs.mu.RUnlock()
 
-	inode, fh := parent.Create(op.Name, op.Metadata)
-
 	parent.mu.Lock()
```
skuppa-veeva, line 1017: The folder creation is moved after acquiring the parent lock. This lock needs to be atomic between creating inode and inserting into map of inodes.

internal/goofys.go
```diff
@@ -1049,14 +1049,14 @@ func (fs *Goofys) MkDir(
...
+	parent.mu.Lock()
+
 	// ignore op.Mode for now
 	inode, err := parent.MkDir(op.Name)
 	if err != nil {
 		return err
 	}
 
-	parent.mu.Lock()
-
```
skuppa-veeva, line 1059: The folder creation is moved after acquiring the parent lock. This lock needs to be atomic between creating inode and inserting into map of inodes.

skuppa-veeva:
@kahing This PR fixes multiple issues:
https://github.com/kahing/goofys/issues/377
https://github.com/kahing/goofys/issues/571
https://github.com/kahing/goofys/issues/353

Could you review and make sure the order of lock is correct and it works as expected.  We are waiting for this fix and I would like to get you feedback.

skuppa-veeva:
@kahing could you review this PR and give us your expert advice.

skuppa-veeva:
@kahing could you review this PR?

monthonk:
This PR seems to fix multiple issues but we need a rebase to make it work with current code base. I'm not sure you're still tracking it but we probably want to try to get this PR merged as part of the next release.

94jskim:
@kahing @gaul @monthonk  I need this fix, when will it be released?

PR #579, opened 2020-11-06
Escape the special char in name correctly.  As per the AWS documentat…

skuppa:
Escape the special char in name correctly.  As per the AWS documentation, CopyRequest is sending request param and it advised to URLEncode the source name.  Using the `url.QueryEscape` to encode the name instead of `url.PahEscape` since it does not cover all special char.

### DIFF
--- a/internal/backend_s3.go
+++ b/internal/backend_s3.go
@@ -475,7 +475,7 @@ func (s *S3Backend) mpuCopyPart(from string, to string, mpuId string, bytes stri
 	params := &s3.UploadPartCopyInput{
 		Bucket:            &s.bucket,
 		Key:               &to,
-		CopySource:        aws.String(pathEscape(from)),
+		CopySource:        aws.String(url.QueryEscape(from)),
 		UploadId:          &mpuId,
 		CopySourceRange:   &bytes,
 		CopySourceIfMatch: srcEtag,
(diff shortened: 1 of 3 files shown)


skuppa:
Fixes #580

internal/backend_s3.go
```diff
@@ -475,7 +475,7 @@ func (s *S3Backend) mpuCopyPart(from string, to string, mpuId string, bytes stri
 	params := &s3.UploadPartCopyInput{
 		Bucket:            &s.bucket,
 		Key:               &to,
-		CopySource:        aws.String(pathEscape(from)),
+		CopySource:        aws.String(url.QueryEscape(from)),
```
skuppa, line 478: Use the `QueryEscape` to encode URLEncode the params.

internal/backend_s3.go
```diff
@@ -667,7 +667,7 @@ func (s *S3Backend) CopyBlob(param *CopyBlobInput) (*CopyBlobOutput, error) {
 
 	params := &s3.CopyObjectInput{
 		Bucket:            &s.bucket,
-		CopySource:        aws.String(pathEscape(from)),
+		CopySource:        aws.String(url.QueryEscape(from)),
```
skuppa, line 670: Use the `QueryEscape` to encode URLEncode the params.  The `from` is a leaf node (just name, not a path) hence `QueryEscape` will escape all special characters.  Even if the `from` has a forward slash (`/`) then it will be escaped and AWS unescaping and working correctly.

internal/goofys.go
```diff
@@ -571,13 +571,6 @@ func mapAwsError(err error) error {
 	}
 }
 
-// note that this is NOT the same as url.PathEscape in golang 1.8,
```
skuppa, line 574: Removed since it is not used anywhere.

kahing:
merged via 5ae08c09108d9e420e77c2ea6b22801be7f46aec

PR #583, opened 2020-11-17
Add a new isAws flag to enable aws v2 api

maobaolong:


maobaolong:
@kahing As Ozone use Goofys as an Office suggestion way to supply fuse program, and now when I use goofys to list a more than 1000 files bucket, goofys will stuck and send listObjectRequest to the Ozone s3g time and time forever unstop.

After a debug, I found that the key point is that goofys use ListObject v1 api to communicate to Ozone s3gateway, but Ozone s3gateway only implemented ListObject v2 version. I tested that if we merge this PR and add a `--isAws` flag can force to use  ListObject v2 version, so this issue can be resolved.

Please take a look, thanks.

kahing:
Could you explain why using ListObjectV1 was causing problems? seems like there's a bug on the server side and you have fixed that?

Issue #590, opened 2020-12-17
Remove (or make optional) default_permissions on Linux.


Arcayr:
Hi there,

I got [a change merged into jacobsa/fuse recently](https://github.com/jacobsa/fuse/pull/94) which adds support for omitting the `default_permissions` option when mounting a filesystem. The original behaviour, which enforced this, was a difference in behaviour between goofys and s3fs-fuse.

Are you interested in making this configurable in any way? It'd be ideal to match the permissions behaviour of s3fs-fuse, and allow containers to export the mounted filesystem as a volume to be shared to other containers.

I'm happy to work on the patch if it's something you're interested in. I already patched my local copy of Goofys and fuse to make this work.

PR #591, opened 2020-12-17
Add --no-default-permissions flag based on latest fuse library.

Arcayr:
As per #590.

This also includes an updated `fuse` library which had some mandatory changes to elsewhere in the codebase.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -15,7 +15,8 @@
 	url = https://github.com/jtolds/gls
 [submodule "vendor/github.com/jacobsa/fuse"]
 	path = vendor/github.com/jacobsa/fuse
-	url = https://github.com/kahing/fusego
+	url = https://github.com/jacobsa/fuse
+	branch = master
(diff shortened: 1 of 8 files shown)


kahing:
any reason to use this instead of say, `--dir-mode=0777 --file-mode=0666`?

Arcayr:
S3 doesn't really have a unix-like permissions model, as far as I'm aware. But mostly for consistency with similar implementations, if I may copy [the commentary](https://github.com/jacobsa/fuse/blob/9677d0392291693bd7b462a199a86f9f226a71be/fuseops/simple_types.go#L67-L88) from the `fuse` library prior to my changes:
```
// Note that in contrast to the defaults for FUSE, this package mounts file
// systems in a manner such that the kernel checks inode permissions in the
// standard posix way. This is implemented by setting the default_permissions
// mount option (cf. http://goo.gl/1LxOop and http://goo.gl/1pTjuk).
```

PR #606, opened 2021-02-15
fix typo

antonpolishko:


README.md
```diff
@@ -6,7 +6,7 @@
 
 # Overview
 
-Goofys allows you to mount an S3 bucket as a filey system.
+Goofys allows you to mount an S3 bucket as a file system.
```
brianpham, line 9: @antonpolishko Its not a typo :) 

```
It's a Filey System instead of a File System because goofys strives for performance first and POSIX second. Particularly things that are difficult to support on S3 or would translate into more than one round-trip would either fail (random writes) or faked (no per-file permission)
```

antonpolishko: is this some kind of insider community meme? The only referral to "filey system" is in this repo :)

PR #612, opened 2021-03-26
Update README.md

xorinox:
Hi, I suggest to mention that for mounting via /etc/fstab fuse needs to be installed. It can be tested via checking for the fuse version: fusermount -V. On my Centos 7.9 sudo yum install -y fuse was needed.


kahing:
it's already mentioned in the readme: https://github.com/kahing/goofys/blame/ab3d914fb8c8d304c2cb31ec21dca80c8e312c23/README.md#L21

PR #668, opened 2021-11-06
Update Homebrew Cask Instructions

knpwrs:
Casks are now installed with `brew install --cask <name>` rather than `brew cask install <name>`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ close-to-open.
 * On macOS, install via [Homebrew](https://brew.sh/):
 
 ```ShellSession
-$ brew cask install osxfuse
+$ brew install --cask osxfuse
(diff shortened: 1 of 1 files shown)


jhult:
`osxfuse` has been succeeded by `macfuse` as of version 4.0.0 (which is why @knpwrs switched this to install `macfuse`).

@kahing, any updates on getting this merged?

dannycjones:
You won't be able to install as a formulae via Homebrew anymore due to Macfuse licensing/distribution (Homebrew/homebrew-core#64491), so this PR wouldn't be ready to merge.

However, we can recommend installing some version of FUSE on macOS and then installing Goofys from elsewhere.

Issue #522, opened 2020-06-04
I/O error with third party


kocane:
So I'm trying to have a third pary tool (Acronis Backup) to include a Goofy mounted S3 bucket in the backup, but it seems to fail. Of course I'm not gonna blame Goofy, but was hoping that maybe I could do something to fix it.

When the backup tool tries to get a hold of the files, I see this error:
```
2020/06/04 16:36:31.515029 fuse.DEBUG Op 0x00000007        connection.go:408] <- OpenDir (inode 1)
2020/06/04 16:36:31.515080 fuse.DEBUG OpenDir 1  []
2020/06/04 16:36:31.515100 fuse.DEBUG Op 0x00000007        connection.go:491] -> OK ()
2020/06/04 16:36:31.515258 fuse.DEBUG Op 0x00000008        connection.go:408] <- ListXattr (inode 1)
2020/06/04 16:36:31.515282 fuse.DEBUG ListXattr 1  []
2020/06/04 16:36:31.515323 s3.DEBUG DEBUG: Validate Request s3/HeadObject failed, not retrying, error InvalidParameter: 1 validation error(s) found.
- minimum field size of 1, HeadObjectInput.Key.

2020/06/04 16:36:31.515337 s3.DEBUG DEBUG: Build Request s3/HeadObject failed, not retrying, error InvalidParameter: 1 validation error(s) found.
- minimum field size of 1, HeadObjectInput.Key.

2020/06/04 16:36:31.515344 s3.DEBUG DEBUG: Sign Request s3/HeadObject failed, not retrying, error InvalidParameter: 1 validation error(s) found.
- minimum field size of 1, HeadObjectInput.Key.

2020/06/04 16:36:31.515358 s3.ERROR code=InvalidParameter msg=1 validation error(s) found., err=InvalidParameter: 1 validation error(s) found.
caused by: ParamMinLenError: minimum field size of 1, HeadObjectInput.Key.

2020/06/04 16:36:31.515369 s3.ERROR code=InvalidParameter msg=1 validation error(s) found., err=InvalidParameter: 1 validation error(s) found.
caused by: ParamMinLenError: minimum field size of 1, HeadObjectInput.Key.

2020/06/04 16:36:31.515385 fuse.DEBUG Op 0x00000008        connection.go:493] -> Error: "InvalidParameter: 1 validation error(s) found.\n- minimum field size of 1, HeadObjectInput.Key.\n"
[...]

PR #672, opened 2022-01-15
Do not try to get xattrs for the root inode

djmaze:
Otherwise the path in the S3 request is empty, which is invalid.

(original PR at https://github.com/ppenguin/goofys/pull/2)


dannycjones:
This resolves kahing/goofys#522.

PR #680, opened 2022-02-28
Update azure dependencies to recent, stable version

### LINKED ISSUE #664: Installation on MacOS using go install fails
Installing source using go...

`go install github.com/kahing/goofys@latest`

ends with the error:
```
# github.com/kahing/goofys/api/common
../../go/pkg/mod/github.com/kahing/goofys@v0.24.0/api/common/conf_azure.go:272:34: accountsRes.Value undefined (type storage.AccountListResultPage has no field or method Value)
../../go/pkg/mod/github.com/kahing/goofys@v0.24.0/api/common/conf_azure.go:373:35: not enough arguments in call to client.ListKeys
	have (context.Context, string, string)
	want (context.Context, string, string, storage.ListKeyExpand)
```

philwinder:
The dependencies listed in the go.mod are pointing to azure libraries that have interfaces that goofys won't work with.

This PR updates the code to use the correct interfaces (I defaulted to empty options, this may not be appropriate) and then I could update the library versions to the most recent.

This fixes #664 and similar build issues.

### DIFF
--- a/api/common/conf_azure.go
+++ b/api/common/conf_azure.go
@@ -269,7 +269,7 @@ func azureFindAccount(client azblob.AccountsClient, account string) (*azblob.End
 		return nil, "", err
 	}
 
-	for _, acc := range *accountsRes.Value {
+	for _, acc := range accountsRes.Values() {
 		if *acc.Name == account {
 			// /subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/...
 			parts := strings.SplitN(*acc.ID, "/", 6)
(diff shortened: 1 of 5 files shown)


dannycjones:
@philwinder do you have a peer that can review and run the tests for these changes? I think its a case of running `make run-tests` with the correct credentials configured. I don't have access to an Azure account nor the context at the moment to review changes for Azure back-end.

I'm hoping we can get this PR reviewed by a member of the community, then final review and merge from Ka-Hing, allowing users to build and install Goofys using `go install`.

dannycjones:
I'm also unable to build this right now - would it be possible for you to rebase to pick up the update to the fuse lib in 829d8e5ce20faa3f9f6f054077a14325e00e9249?

Currently seeing the following error, which seems related to the FUSE lib updated in that commit:

```
$ make build
go build -ldflags "-X main.Version=`git rev-parse HEAD`"
# github.com/kahing/goofys
github.com/jacobsa/fuse/internal/buffer.memclr: relocation target runtime.memclrNoHeapPointers not defined for ABI0 (but is defined for ABIInternal)
make: *** [Makefile:13: build] Error 2
```

kahing:
closing via cc76eec5841f7bbf499e21d97f91a01e2e2f5af0

Issue #696, opened 2022-05-19
Update Readme needed. Go older than 1.15 or newer than 1.16 does not work with installation 


rezuma:
The code and installation instructions does not support any other version than Go 1.16
I try with Go 1.13 and get
# github.com/golang-jwt/jwt
go/src/github.com/golang-jwt/jwt/ecdsa.go:135:4: r.FillBytes undefined (type *big.Int has no field or method FillBytes)
go/src/github.com/golang-jwt/jwt/ecdsa.go:136:4: s.FillBytes undefined (type *big.Int has no field or method FillBytes)

Fillbytes support started in go 1.15

G 1.17 deprecated go get so instead you have to use go install github.com/kahing/goofys@latest

After I tried installing it with Go 1.18 then i get this already reported error

# github.com/kahing/goofys/api/common
work/pkg/mod/github.com/kahing/goofys@v0.24.0/api/common/conf_azure.go:272:35: accountsRes.Value undefined (type storage.AccountListResultPage has no field or method Value)
work/pkg/mod/github.com/kahing/goofys@v0.24.0/api/common/conf_azure.go:373:67: not enough arguments in call to client.ListKeys
	have (context.Context, string, string)
	want (context.Context, string, string, storage.ListKeyExpand)

dannycjones:
We should update this line with the minimum Go version supported.

https://github.com/kahing/goofys/blob/master/go.mod#L3

PR #687, opened 2022-03-31
brew cask disabled

with-kim:
Error: `brew cask` is no longer a `brew` command. Use `brew <command> --cask` instead.

- brew cask commands were deprecated on 2020-12-01 with the release of Homebrew 2.6.0. Starting then, all brew cask commands succeeded but displayed a warning informing users that the command would soon be disabled. The message also provides the appropriate replacement.
- brew cask commands were disabled on 2020-12-21 with the release of Homebrew 2.7.0. Starting then, all brew cask commands failed and displayed a warning informing users that the command is disabled. The message also provides the appropriate replacement.
- With the release of Homebrew 2.8.0 (release date TBD), this disable message will be removed.

Issue #628, opened 2021-05-18
Failed to install goofys on MacOS


KimKiHyuk:
## Enviroment
ProductName:    Mac OS X
ProductVersion: 10.15.7
BuildVersion:   19H524

I tried to install goofys via homebrew 

```
$ brew install --cask osxfuse
$ brew install goofys
```
when running after `brew install goofys`, I've encountered a message that
```
Error: goofys has been disabled because it requires FUSE!
```

KimKiHyuk:
oh, I think it's homebrew problem. some features have been changed recently

https://stackoverflow.com/questions/67155330/installing-ifuse-with-homebrew-results-in-error-message

I'll change mac os installation document after fix it

wsong1106:
Hi! The same issue happens to me, just want to check with you do you solve this problem?

joeyparis:
Update as of 2021-09-19, same error message but install process has changed. Now `brew install macfuse` instead of `brew install [--cask] osxfuse`, but still has the same error at the end

jhult:
Related to #603 and https://github.com/kahing/goofys/pull/668.

PR #700, opened 2022-05-25
Update instructions for new macOS guidance and drop source install for now

dannycjones:
In summary, this PR reworks the installation instructions in two ways.

Firstly, it drops the installation from source instructions, with the intention to add an updated version back once working. As #696 mentions, `go install` should be used instead of `go get` now. In terms of getting `go install` working, I'm seeing the same issues as #527, #664.

Secondly, it updates the instructions now that goofys is not available from Homebrew. This is due to a licensing change on the FUSE library used on macOS. Addresses #628.

Note, this does overlap with #618, #668, #687 but they do not address that goofys can no longer be installed via Homebrew (see Homebrew/homebrew-core#64491).

(closes #618, #668, #687)

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,14 +27,6 @@ $ brew cask install osxfuse
 $ brew install goofys
 ```
 
-* Or build from source with Go 1.10 or later:
-
-```ShellSession
-$ export GOPATH=$HOME/work
-$ go get github.com/kahing/goofys
-$ go install github.com/kahing/goofys
-```
-
 # Usage
 
 ```ShellSession
--- a/README.md
+++ b/README.md
@@ -18,14 +18,18 @@ close-to-open.
 
 # Installation
 
-* On Linux, install via [pre-built binaries](https://github.com/kahing/goofys/releases/latest/download/goofys). You may also need to install fuse-utils first.
+Ensure you have FUSE available on your system.
 
-* On macOS, install via [Homebrew](https://brew.sh/):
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -18,22 +18,18 @@ close-to-open.
...
-* Or build from source with Go 1.10 or later:
+## Pre-built binaries
 
-```ShellSession
-$ export GOPATH=$HOME/work
-$ go get github.com/kahing/goofys
-$ go install github.com/kahing/goofys
-```
+You can [download pre-built goofys binaries](https://github.com/kahing/goofys/releases/latest/download/goofys) available on GitHub.
+Move these to a location in your path if you prefer to run without specifying a path to goofys.
```
monthonk, line 32: I think we should use upper case PATH here, so we know it means PATH variable.

monthonk (changes requested):
Suggested a little change but everything else looks good for now.

We really need a new release so we will have `go install` working.

README.md
dannycjones: Sounds good, updated in fcf132c

monthonk (approved):

monthonk:
Looks good, thanks Danny.

dannycjones:
If #680 is merged before this, I think we can re-introduce the building from source options subject to testing.

dannycjones:
Closing. Change is outdated and Homebrew is no longer a supported method for install FUSE dependent connectors on macOS.

Issue #707, opened 2022-07-10
Panic / ForgetInodeOp error: input/output error


cryptowhizzard:
[...]

PR #721, opened 2022-09-26
fix issue:Panic / ForgetInodeOp error: input/output error #707

ifoxhz:
Lock the inode by fs.mu


kahing:
I think that's an outdated comment, since we already use `atomic` package when using `refcnt`. It's not clear the PR would change anything

Issue #732, opened 2022-11-16
Failed to compile from source neither on arm64 or x86


ddynwzh1992:
Hi,

I am trying to compile the goofys on Graviton2, as there are no arm64 package available.
However when I follow the steps to build from source, following error messages shown up

# github.com/kahing/goofys/api
api/api.go:124:56: cannot use common.FusePanicLogger{...} (type common.FusePanicLogger) as type fuseutil.FileSystem in argument to fuseutil.NewFileSystemServer:
	common.FusePanicLogger does not implement fuseutil.FileSystem (missing BatchForget method)

Then I try to use x86 instance, but failed as well, with same error message 

I also have tried different go version including 1.17, 1.18, 1.19 but no lucky 

Please help to verify this

Thanks,
Vincent

monthonk:
Hi, this PR https://github.com/kahing/goofys/pull/733 should fix the compiling issue.
At least it works on x86, would you be able to try it on arm64 as well?
thanks

ddynwzh1992:
It works! Both x86 and arm64.
Thanks a lot

whls:
sorry, I forget to change the file api/common/panic_logger.go.
After changed, I compiled successful!
Thanks!

PR #733, opened 2022-11-16
Fix build error from dependency update

monthonk:
Fix build error after updating jacobsa/fuse to v0.0.0-20221016084658-a4cd154343d8 (https://github.com/kahing/goofys/issues/732)

### DIFF
--- a/api/common/panic_logger.go
+++ b/api/common/panic_logger.go
@@ -60,6 +60,10 @@ func (fs FusePanicLogger) ForgetInode(ctx context.Context, op *fuseops.ForgetIno
 	defer LogPanic(&err)
 	return fs.Fs.ForgetInode(ctx, op)
 }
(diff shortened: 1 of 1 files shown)


api/api.go
```diff
@@ -121,7 +121,7 @@ func Mount(
 		err = fmt.Errorf("Mount: initialization failed")
 		return
 	}
-	server := fuseutil.NewFileSystemServer(FusePanicLogger{fs})
+	server := fuseutil.NewFileSystemServer(FusePanicLogger{Fs: fs})
```
gaul, line 124: Is this needed?

monthonk: no, it's not.

I just got a warning `struct literal uses unkeyed fields` in my IDE and though we should address it.

monthonk: I'm happy to revert this file if that's a better choice

gaul (approved):

PR #761, opened 2023-06-17
Remove duplicate definition of BatchForget function

mechanicker:
This fixes build on both Linux & MacOS with Go 1.20


mechanicker:
This is a duplicate

Issue #230, opened 2017-09-26
AutoFS hangs


johanot:
Trying to mount bucket with goofys using autofs, but..
When I try to cd into the mount location, the terminal process hangs and mounting obviosly doesn't work:

Trying to access autofs-mounted bucket:
`[ec2-user@ip-192-168-1-81 ~]$ cd /data/app
^C^C-bash: cd: /data/app: Interrupted system call`

Content of autofs master map (/etc/auto.master)
`/data  /etc/auto.data  --ghost --timeout=10`

/etc/auto.data
`app    -fstype=fuse,ro,_netdev,allow_other,--region=eu-central-1,--debug_fuse,--debug_s3,--file-mode=0666  :/usr/bin/goofys\#bucket`

/var/log/messages output

 26 14:17:45 ip-192-168-1-81 automount[29215]: mounted indirect on /data with timeout 300, freq 75 seconds
Sep 26 14:17:45 ip-192-168-1-81 automount[29215]: st_ready: st_ready(): state = 0 path /data
Sep 26 14:18:35 ip-192-168-1-81 automount[29215]: handle_packet: type = 3
Sep 26 14:18:35 ip-192-168-1-81 automount[29215]: handle_packet_missing_indirect: token 172, name app, request pid 18912
Sep 26 14:18:35 ip-192-168-1-81 automount[29215]: attempting to mount entry /data/app
Sep 26 14:18:35 ip-192-168-1-81 automount[29215]: lookup_mount: lookup(file): looking up app
Sep 26 14:18:35 ip-192-168-1-81 automount[29215]: lookup_mount: lookup(file): app -> -fstype=fuse,ro,_netdev,allow_other,--region=eu-central-1,--debug_fuse,--debug_s3,--file-mode=0666  :/usr/bin/goofys\#bucket
Sep 26 14:18:35 ip-192-168-1-81 automount[29215]: parse_mount: parse(sun): expanded entry: -fstype=fuse,ro,_netdev,allow_other,--region=eu-central-1,--debug_fuse,--debug_s3,--file-mode=0666  :/usr/bin/goofys\#bucket
Sep 26 14:18:35 ip-192-168-1-81 automount[29215]: parse_mount: parse(sun): gathered options: fstype=fuse,ro,_netdev,allow_other,--region=eu-central-1,--debug_fuse,--debug_s3,--file-mode=0666
[...]

PR #763, opened 2023-06-25
Signal parent once child takes over

mechanicker:
Fix for: https://github.com/kahing/goofys/issues/230

Parent process was blocked waiting on `wg` sync.WaitGroup preventing autofs from managing the forked daemon child process. Sending the signal to parent once child is ready ensures parent process exits and lets child process handle mounting and serving FUSE requests.

Tested this in an EC2 instance with existing autofs (for NFS) with `goofys`:
```
$ cat /etc/auto.master
/mnt/data  /etc/auto.nfs
/mnt/data/s3  /etc/auto.s3

$ cat /etc/auto.s3
data -fstype=fuse,_netdev,allow_other,--region=us-west-2 :goofys#s3-bucket-name

# List contents of git repository in S3 bucket
$ ls /mnt/data/s3/data/repos/puntar/
HEAD  config  description  hooks  info  objects  packed-refs
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -196,6 +196,9 @@ func main() {
 				kill(os.Getpid(), syscall.SIGUSR1)
 				wg.Wait()
 				defer ctx.Release()
+
+				// Signal parent process since we are taking over
+				_ = kill(os.Getppid(), syscall.SIGUSR1)
 			}
 
 		} else {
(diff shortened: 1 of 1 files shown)


mechanicker:
@kahing Could you please help review and merge this important fix to get `goofys` working with automounter/autofs?

rockeet:
```go
defer func() {
         time.Sleep(time.Second)
         flags.Cleanup()
}()
```
@mechanicker  The defere func sleep 1 second which is too long in our use case, when we delete the defer func(leak flags?), it is ok for old code, but with this PR, parent process exit immediately while the mount operation have not done.

Could you provide a fix to exit parent exactly after child mounted?

rockeet:
#783 Fixed this issue.

PR #778, opened 2023-12-19
chore: Elevate the gopsutil module to v3

chiehting:
Upgrade the gopsutil module to facilitate the correct compilation of the goofys binary on macOS M1.


martincpt:
Would love to see this getting merged. Probably closes #798

Colelyman:
I agree with @martincpt, for anyone else that stumbles here, I was able to get it running on M1 Mac 12.7.5 after checking out this branch and running `go build`. The `brew` install failed as well as installing v0.24.0.

PR #786, opened 2024-01-11
Fix ReadDir inode leak

### LINKED ISSUE #785: ReadDir inode leak causes memory leak
# What happened?
The memory of the goofys process keeps growing and cannot reclaim the memory via
`sync;echo 3 > /proc/sys/vm/drop_caches;killall -USR1 goofys;`
Use go pprof heap to analyze memory
image
The inode created by ReadDir takes up a lot of memory
# How can we reproduce it
I wrote a simple go program to reproduce the problem
``` go
package main

import (
	"fmt"
	"os"
)

func main() {
	_, err := os.ReadDir(os.Args[1])
	if err != nil {
		fmt.Println(err)
		return
	}
	fmt.Println(os.Args[1])
}
```
Create 1000 files in the /mnt/goofys/test directory
Repeat run
``` bash
go run main.go /mnt/goofys/test
sync;echo 3 > /proc/sys/vm/drop_caches;killall -USR1 goofys;
```
goofys log
```
/root/goofys: main.INFO Received user defined signal 1
/root/goofys: main.INFO forgot xxx inodes
/root/goofys: main.INFO 1004 inodes

/root/goofys: main.INFO Received user defined signal 1
/root/goofys: main.INFO forgot xxx inodes
/root/goofys: main.INFO 2003 inodes

/root/goofys: main.INFO Received user defined signal 1
/root/goofys: main.INFO forgot xxx inodes
/root/goofys: main.INFO 3006 inodes
...
```
The number of unrecyclable inodes will continue to increase
# Problem causes
goofys ReadDir will cache all child inodes, but OS only caches the inodes of the parent directory. When memory is tight or `echo 3 > drop_caches`. OS call goofys to forget the parent inode. At this time, all child nodes will become dirty data exist in The fs.inodes map cannot be recycled.
# What did you expect to happen?
`sync;echo 3 > /proc/sys/vm/drop_caches;killall -USR1 goofys;`can reclaim the memory

1978629634:
fix https://github.com/kahing/goofys/issues/785
When the OS reclaims the inode of the parent directory, delete the child node from the global map

### DIFF
--- a/internal/goofys.go
+++ b/internal/goofys.go
@@ -764,6 +764,16 @@ func (fs *Goofys) ForgetInode(
 		fs.mu.Lock()
 		defer fs.mu.Unlock()
 
+		if inode.isDir() {
+			for _, child := range inode.dir.Children {
+				if *child.Name == "." || *child.Name == ".." {
+					continue
+				}
+				delete(fs.inodes, child.Id)
+				fs.forgotCnt += 1
+			}
+		}
+
 		delete(fs.inodes, op.Inode)
 		fs.forgotCnt += 1
 


1978629634:
@kahing Excuse me, could you help me review it, please?

fenglangwangyi:
Is the issue resolved?

